Issue #22, opened 2015-02-23
Exception is raised when init_app() is called more than once


croshchupkin:
The culprit is this line of code in init_app():
```
self.add_resource(self.swagger_view(), '/swagger.json', endpoint='specs', doc=False)
```
The self.swagger_view() method returns a new class each time. And, because of that, the code in the following if-block gets executed: https://github.com/flask-restful/flask-restful/blob/0.3.1/flask_restful/__init__.py#L391

This is very inconvenient for me, as because of this issue I can't create more than one class for my functional tests: for each such class, I must call my application factory, which, in turn, imports my Api instance and calls init_app() on it.

The fastest solution that I see right now is to just make the addition of the specs optional via a new kwarg. I can create a pull request for this.

croshchupkin:
I've added 2 tests for the new feature to the pull request.

noirbizarre:
Fixed by #23

croshchupkin:
Could you please create a new version that contains these changes?

noirbizarre:
I will very soon. (Today or tomorrow)

PR #23, opened 2015-02-23
#22 optional specs endpoint addition

croshchupkin:
#22


noirbizarre:
Hi !

Thanks for this pull-request, it might be very usefull.
Do you mind adding the corresponding tests before I merge it ?

croshchupkin:
Ok, I will try to. But it'll most likely happen on Saturday or Sunday.

croshchupkin:
Sorry for the delay, here are 2 additional tests. I couldn't really think of anything else to test which involved the `add_specs` parameter (the tests for the presense of /swagger.json are virtually, if somewhat implicitly, present in test_swagger).

PR #40, opened 2015-04-12
Fix typo

awiddersheim:
I might be wrong on this one but the example for polymorphism doesn't look right.


noirbizarre:
Your're right !

PR #69, opened 2015-09-07
Fix incorrect example

awiddersheim:
I could be wrong about this example being incorrect but judging from the other parts of the documentation about the security stuff this seems right.

### DIFF
--- a/doc/documenting.rst
+++ b/doc/documenting.rst
@@ -712,7 +712,7 @@ You can apply this requirement globally with the security constructor parameter:
             'keyname': HEADER_API_KEY
         }
     }
-    api = Api(app, authorizations=authorizations, security='security')
+    api = Api(app, authorizations=authorizations, security='apikey')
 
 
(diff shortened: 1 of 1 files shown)


noirbizarre:
No, this is example is correct. You have to pass as security (on either the API or a method) one key from security definitions (was authorizations before, this why is called authorizations in API constructor).
Flask-restplus allow to pass a single string or a list of string or object ...

awiddersheim:
Yes, but I don't see `security` as a key in the `authorizations` definition you are declaring just before:

```
     authorizations = {
         'apikey': {
             'type': 'apiKey',
             'passAs': 'header',
             'keyname': HEADER_API_KEY
         }
     }
api = Api(app, authorizations=authorizations, security='security')
```

noirbizarre:
Sorry, you were right. I read the diff in the wrong direction :)

awiddersheim:
No worries. I do the same thing all the time.

PR #75, opened 2015-09-26
Use form data when validating if json not present

awiddersheim:
In lieu of proper content negotiation use the form data from the request
if no json data was present.

### DIFF
--- a/flask_restplus/resource.py
+++ b/flask_restplus/resource.py
@@ -16,8 +16,8 @@ class Resource(restful.Resource):
     def validate_payload(self, func):
         '''Perform a payload validation on expected model'''
         def wrapper(*args, **kwargs):
-            # TODO: proper content negociation
-            data = request.get_json()
+            # TODO: proper content negotiation
+            data = request.get_json() or request.form
             if hasattr(func, '__apidoc__'):
                 model = func.__apidoc__.get('body')
                 validate = func.__apidoc__.get('validate', False)


noirbizarre:
Thanks for this PR.

Can I have a testcase for this one ?

awiddersheim:
Yes, sorry. Let me whip something together.

awiddersheim:
Looks like this doesn't work as I anticipated. Types aren't maintained when getting data from forms it seems.

awiddersheim:
When giving the same data as in the `test_api_payload` test:

```
data = {
    'name': 'John Doe',
    'age': 15,
}
```

The `request.form` property returns the following:

`ImmutableMultiDict([('age', u'15'), ('name', u'John Doe')])`

Which fails to validate because of the following:

`{"message": "Input payload validation failed", "errors": {"age": "u'15' is not of type u'integer'"}}`

awiddersheim:
No obvious way to fix that I can see without adding proper content negotiation and different validation methods for each. Will leave open for @noirbizarre to review and close at his discretion.

noirbizarre:
Hum, it seem that more than validation, we should cast and transform before.
Not a trivial one.

What I propose to you: first I release the 0.8.0 and we work on that for the 0.8.1.
Is it fine to you ?

ps: this is why I ask for tests ;)

awiddersheim:
Sounds good to me. Yeah, I meant to add them not sure why I didn't. I had this working but only with strings as the input. Wasn't until I added an integer in the test that I had issues.

PR #87, opened 2015-10-26
Force path value to be of type str

l-vincent-l:


l-vincent-l:
You cannot join a int, and this can happen when you have a list

PR #96, opened 2015-11-20
Hide resource if doc is False

AleixDev:


noirbizarre:
Hum, not sure what you are trying to achieve.
The current code state allow to write `@api.doc(False)` to ignore documentation.

Maybe I misunderstand but withtout details nor test, I can't merge this.

AleixDev:
Yes it's working fine with:
`@api.doc(False)`

But I was trying to do this (as explained in the [documentation][1]):
`@api.route('/resource1/', doc=False)`

[1]: http://flask-restplus.readthedocs.org/en/stable/documenting.html#hiding-from-documentation

noirbizarre:
OK, I just understood your PR. You are trying to do `@ns.route(doc=False)` which is not tested.
I was trying to figure out where you code came from because it's already implemented an api:
https://github.com/noirbizarre/flask-restplus/blob/master/flask_restplus/api.py#L227

I cherry-picked your commit and added a test for it.

AleixDev:
Yes, I'm using a namespace, not the root api. I forgot to mention it, thank you for taking the time to understand this PR :)

PR #99, opened 2015-11-27
Fixes #22: Exception is raised when init_app() is called more than once

### LINKED ISSUE #22: Exception is raised when init_app() is called more than once

frol:
Original issue is #22. Fix in #23 was not complete.

Once `add_resource` is called, you cannot undo it and every next run of `init_app` with another Blueprint or Flask-app will inevitably augment them with `/swagger.json` endpoint even if `add_specs=False`. Even worse, if `init_app` is called with `add_specs=True` (default) the second time, it would fail as the specs resource was already added previously.

My fix is cleaner as it mounts specs resource straight to the Blueprint/Flask-app, just like `_register_doc` does.

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -175,13 +175,12 @@ class Api(restful.Api):
         self.contact_email = kwargs.get('contact_email', self.contact_email)
         self.license = kwargs.get('license', self.license)
         self.license_url = kwargs.get('license_url', self.license_url)
-
-        if kwargs.get('add_specs', True):
-            self.add_resource(self.swagger_view(), '/swagger.json', endpoint='specs', doc=False)
+        self._add_specs = kwargs.get('add_specs', True)
 
         super(Api, self).init_app(app)
 
     def _init_app(self, app):
+        self._register_specs(self.blueprint or app)
         self._register_doc(self.blueprint or app)
         super(Api, self)._init_app(app)
         self._register_apidoc(app)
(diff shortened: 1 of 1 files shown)


frol:
@noirbizarre Tests are passed now (small Python 2.x regression was fixed), please, review.

flask_restplus/api.py
```diff
@@ -207,8 +216,6 @@ def documentation(self, func):
 
     def swagger_view(self):
```
noirbizarre: You're right, the factory function is not necessray anymore.
We can extract SwaggerView somewhere (make it more reuseable) and call:
```python
self._register_view(
    app_or_blueprint,
    SwaggerView,
    '/swagger.json',
    endpoint=str('specs'),
    resource_class_args=(self, )
)
```

noirbizarre:
I'l do the SwaggerView changer myself

PR #100, opened 2015-11-27
Keep order in parameters.

AleixDev:
This should keep the order of parameters when doing something like:
```
parser.add_argument('first_example', type=str)
parser.add_argument('second_example', type=str)
parser.add_argument('third_example', type=str)
```

Not ordered:
not_ordered

Ordered:
ordered


AleixDev:
Thanks!

Issue #103, opened 2015-12-02
Swagger documentation breaks when Blueprint has subdomain


sharan-karanth:
I am currently on version 0.8.2 of your wonderful API.

When I register my API blueprint with a subdomain like in the code below

```
api_bp = Blueprint('api', __name__, subdomain="api",  url_prefix='/api')
flask_api = Api(api_bp,
                  version='1.0-Beta',
                  title='My API',
                  description='Some description',
                  default='General',
                  default_label='General Operations')

```
The generated swagger.json has an invalid host url.
The culprit code is in swagger.py, line 185
```
   'host': current_app.config.get('SERVER_NAME', None) or None,
```

It does not take into consideration the subdomain. Leaving it as None works 

One option is to leave it as none if api.blueprint.subdomain is not None.
Another option would be to allow setting of host url in Api init and use that if provided.

sharan-karanth:
Thank you

noirbizarre:
You're welcome ;)

Issue #41, opened 2015-04-14
Problem with PUT method when the param field has the same name than a field in the model


Seraf:
Hello,

I decided to use name instead of id as argument to find my role.
I updated it like this : https://gist.github.com/Seraf/23d0d34e11c38f922f4b

The get(name) and delete(name) works great. But the PUT method has a problem : 
http://127.0.0.1:5000/api/1/core/role/{name} -> the name here isn't replaced by the name value I give
On the get function, it uses the name I give : 
http://127.0.0.1:5000/api/1/core/role/test

I think the problem is that PUT method use the name of my model and not the param field in my decorator.
I think I can workaround the problem using "rolename" as param but it won't make sense on my PUT method to have rolename param and name param with the same value.
I hope I'm clear in my explanation :)

Thanks !

PR #104, opened 2015-12-02
Added ability to override host_url when setting the value for host in…

sharan-karanth:
This fix will allow the client to pass a host url that will be used instead of app.config["SERVER_NAME"] if provided.

Let me know what you think

Sharan

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -118,7 +118,7 @@ class Api(restful.Api):
             contact=None, contact_url=None, contact_email=None,
             authorizations=None, security=None, doc='/', default_id=default_id,
             default='default', default_label='Default namespace', validate=None,
-            tags=None, **kwargs):
+            tags=None, host_url=None, **kwargs):
         self.version = version
         self.title = title or 'API'
         self.description = description
@@ -136,6 +136,7 @@ class Api(restful.Api):
         self._doc_view = None
         self._default_error_handler = None
         self.tags = tags or []
+        self._host_url = host_url
 
         self._error_handlers = {
             ParseError: mask_error_handler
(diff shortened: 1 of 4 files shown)


noirbizarre:
Hi !

Thanks for the contrib, but before merging I need to understand the use case and to discuss if it's the better way of doing it.

With this PR, is the point is being able to give a sandbox URL as host URL in the swagger.json specification (and only in the specifications) ? Is this for everyone who fetch/use the specifications or only for SwaggerUI ?

My point is that if this is only for swagger-ui to use a sanbox instead of the real API, maybe it would be better to implement this at query time with a query string. (ie. swagger.json?host_url=test.api.com) allowing to specify host_url only in your swagger-ui instance.
Some user may want to use your specifications for real use and it will be misleading.

I was working on fixing #41 by properly using the subdomain in blueprint (I need a little bit of time because I never used the subdomain feature it).
But I want to be sure to understand that this is really another use case.

sharan-karanth:
This is a fix for the issue described in issue #103 

In my case, the api will always be in its own subdomain (and Blueprint). And these subdomains will exist in all environments (staging, production).

So, while your recommendation of using host_url as a query parameter would work, I would never want the swagger page to show any other server name as the host since I am initializing the Api with my Blueprint.

Does that make sense?

noirbizarre:
Sorry, I was talking about #103 too.

For me the proper fix would be to really use the blueprint subdomain value instead of having to manually set `host_url`.
subdomain can be dynamic, so the top of the top would be to be able to detect the current subdomain for dynamic subdomains

sharan-karanth:
Yeah. You are right. I had suggested both but went with the latter because I could get it out quicker. Will send you a pull request for this later today.

Thanks
Sharan

noirbizarre:
Can I close this PR as #103 should be fixed?

PR #107, opened 2015-12-07
Add ability to pass headers from error handlers

awiddersheim:
Give user ability to pass headers from default or custom error handlers. There may be better ways to do this so I am open to suggestions. Perhaps using a more unique key like `__headers` or `__restplus_headers` as to avoid conflicting with anything. Another alternative I can think of is to make the exceptions return a third tuple value that contains the headers.

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -29,7 +29,7 @@ from .namespace import Namespace
 from .postman import PostmanCollectionV1
 from .resource import Resource
 from .swagger import Swagger
-from .utils import merge, default_id, camel_to_dash
+from .utils import merge, default_id, camel_to_dash, unpack
 from .reqparse import RequestParser
 
 RE_RULES = re.compile('(<.*>)')
@@ -462,10 +462,10 @@ class Api(restful.Api):
         elif e.__class__ in self._error_handlers:
             handler = self._error_handlers[e.__class__]
             result = handler(e)
-            default_data, code = result if len(result) == 2 else (result, 500)
+            default_data, code, headers = unpack(result, 500)
         elif self._default_error_handler:
             result = self._default_error_handler(e)
-            default_data, code = result if len(result) == 2 else (result, 500)
+            default_data, code, headers = unpack(result, 500)
         else:
             code = 500
             default_data = {
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -144,6 +144,32 @@ class APITestCase(TestCase):
                 'test': 'value',
             })
 
+    def test_errorhandler_for_custom_exception_with_headers(self):
+        api = restplus.Api(self.app)
+
+        class CustomException(RuntimeError):
+            pass
+
+        @api.route('/test/', endpoint='test')
+        class TestResource(restplus.Resource):
+            def get(self):
+                raise CustomException('error')
+
+        @api.errorhandler(CustomException)
+        def handle_custom_exception(error):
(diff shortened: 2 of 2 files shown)


noirbizarre:
Thanks, for the contribution.
I like the feature, but to stay homogeneous with thez flask-restplus behavior, I'd prefer to use the 3-tuple style response. I'll comment your code ASAP for you to see.

tests/test_errors.py
```diff
@@ -125,6 +125,32 @@ def handle_custom_exception(error):
...
+            pass
+
+        @api.route('/test/', endpoint='test')
+        class TestResource(restplus.Resource):
+            def get(self):
+                raise CustomException('error')
+
+        @api.errorhandler(CustomException)
+        def handle_custom_exception(error):
+            return {'message': 'some maintenance', 'headers': {'Retry-After': 120}}, 503
```
noirbizarre: Using the same style as Resources methods response would be better because homogeneous:
```python
@api.errorhandler(CustomException)
def handle_custom_exception(error):
    return {'message': 'some maintenance'}, 503, {'Retry-After': 120}
```

tests/test_errors.py
```diff
@@ -184,6 +210,29 @@ def default_error_handler(error):
...
+        api = restplus.Api(self.app)
+
+        @api.route('/test/', endpoint='test')
+        class TestResource(restplus.Resource):
+            def get(self):
+                raise Exception('error')
+
+        @api.errorhandler
+        def default_error_handler(error):
+            return {'message': 'some maintenance', 'headers': {'Retry-After': 120}}, 503
```
noirbizarre: same as above

flask_restplus/api.py
```diff
@@ -492,6 +492,7 @@ def handle_error(self, e):
             }
 
         default_data['message'] = default_data.get('message', str(e))
+        headers = default_data.pop("headers", headers)
```
noirbizarre: It wouyld be better to use  ```flask_restful.unpack``` (will be merged soon) to properly unpack response above:
```python
elif e.__class__ in self._error_handlers:
    handler = self._error_handlers[e.__class__]
    result = handler(e)
    default_data, code, headers = restful.unpack(result)
elif self._default_error_handler:
    result = self._default_error_handler(e)
    default_data, code, headers = restful.unpack(result)
```
The only problem is that `unpack` always take 200 as default status code.
If you wait a little bit, I'l merge `unpack` allowing to override the default status code and to write:
```python
if isinstance(e, HTTPException):
    code = e.code
    default_data = {
        'message': getattr(e, 'description', HTTP_STATUS_CODES.get(code, ''))
     }
     headers = e.get_response().headers
elif e.__class__ in self._error_handlers:
    handler = self._error_handlers[e.__class__]
    result = handler(e)
    default_data, code, headers = restful.unpack(result, 500)
elif self._default_error_handler:
    result = self._default_error_handler(e)
    default_data, code, headers = restful.unpack(result, 500)
else:
    code = 500
    default_data = {
        'message': HTTP_STATUS_CODES.get(code, str(e)),
    }
    headers = Headers()
```

What do you think about it ?

awiddersheim: Looks great. I'll wait for `unpack` and rebase my branch from `master` and implement the changes.

noirbizarre: It's done. You can rebase and implement it :wink:

awiddersheim:
I rebased off of master and implemented things using `unpack`.

noirbizarre:
Thanks !

PR #118, opened 2016-01-12
Fix resource doc hiding

awiddersheim:
Fix resource and method documentation hiding when doing @api.hide or
@api.doc(False).

### DIFF
--- a/flask_restplus/resource.py
+++ b/flask_restplus/resource.py
@@ -57,7 +57,7 @@ class Resource(MethodView):
 
     def validate_payload(self, func):
         '''Perform a payload validation on expected model if necessary'''
-        if hasattr(func, '__apidoc__'):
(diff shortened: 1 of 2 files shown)


flask_restplus/resource.py
```diff
@@ -57,7 +57,7 @@ def dispatch_request(self, *args, **kwargs):
 
     def validate_payload(self, func):
         '''Perform a payload validation on expected model if necessary'''
-        if hasattr(func, '__apidoc__'):
+        if hasattr(func, '__apidoc__') and func.__apidoc__ != False:
```
noirbizarre: Better with:
```python
if getattr(func, '__apidoc__', False) != False:
```

awiddersheim: Yes, good call. Updated.

PR #121, opened 2016-01-22
Add documentation for serve_challenge_on_401

bodgit:
I was trying to debug why the Swagger UI wasn't making the browser prompt for basic authentication and eventually found the `serve_challenge_on_401` option which was added with the dropping of Flask-Restful as a requirement but it doesn't show up in the generated documentation. So here's a small PR that should add it and prevent anyone else scratching their head :smile:


noirbizarre:
Thanks ;)

Issue #122, opened 2016-01-24
Custom prefix does not work well.


leiserfg:
If api is created with a custom prefix, the doc work bad (the requests from the doc don't take prefix into acount).

leiserfg:
I was reading the source and I think the problem will be fixed if root endpoint url are prefix and not '/'.

frol:
This can be closed since the PR was merged.

PR #123, opened 2016-01-24
Change root endpoint url to prefix if there is one

leiserfg:
It fix the issue #122


noirbizarre:
Hi !
Thanks for the contribution. I'll just need a test case here https://github.com/noirbizarre/flask-restplus/blob/master/tests/test_apidoc.py before merging this one.

leiserfg:
Done!

noirbizarre:
Thanks !

Issue #124, opened 2016-01-24
Cannot use fields.String enum properly


liiight:
I try to use req parser with a fields.String enum to display a nice dropdown in swagger:

```python
seen_search_parser = api.parser()
seen_search_parser.add_argument('page', type=int, default=1, help='Page number')
seen_search_parser.add_argument('max', type=int, default=100, help='Seen entries per page')
seen_search_parser.add_argument('local_seen', type=fields.String(enum=['true', 'false', 'all'], default='all'),
                                help='Filter list by local status.')
```

I the use the `expect` wrapper and I do see this nicely in swagger. But when I try to send the request I get a validation error. When debugging I get a `error: String object is not callable` raised from the `reqparse.convert` method.

Why is that? I am sending a string value obviously under the parameter. What am I doing wrong?

noirbizarre:
Hi!

`RequestParser` expect types from `flask_restplus.inputs` module. (See https://github.com/noirbizarre/flask-restplus/blob/master/flask_restplus/inputs.py#L372-L393 for example)

You can write your own, in you case:
```python
def local_seen(value):
    if value not in ('true', 'false', 'all'):
        raise ValueError('Unexpected value {0}'.format(value))

seen_search_parser = api.parser()
seen_search_parser.add_argument('page', type=int, default=1, help='Page number')
seen_search_parser.add_argument('max', type=int, default=100, help='Seen entries per page')
seen_search_parser.add_argument('local_seen', type=local_seen, default='all',
                                help='Filter list by local status.')


class MyResource(Resource):
    @api.expect(seen_search)
    def get(self):
         args = seen_search..parse_args()
         local_seen = args['local_seen']
         if local_seen == 'true':
             # true case
         elif local_seen == 'false':
             # false case
         elif local_seen == 'all':
             # all case
```

My advice, use boolean and assume 'all' when not specified:
```python
from flask_restplus.inputs import boolean
seen_search_parser = api.parser()
seen_search_parser.add_argument('page', type=int, default=1, help='Page number')
seen_search_parser.add_argument('max', type=int, default=100, help='Seen entries per page')
seen_search_parser.add_argument('local_seen', type=boolean, help='Filter list by local status.')

class MyResource(Resource):
    @api.expect(seen_search)
    def get(self):
         args = seen_search..parse_args()
         local_seen = args['local_seen']
         if local_seen is True:
             # true case
         elif local_seen is False:
             # false case
         elif local_seen is None:
             # all case
```

[...]

PR #126, opened 2016-01-27
Added support for bool

stevezau:
Added bool support for args


noirbizarre:
There is already `flask_restplus.inputs.boolean` (https://github.com/noirbizarre/flask-restplus/blob/master/flask_restplus/inputs.py#L372-L395).

See: https://github.com/noirbizarre/flask-restplus/issues/124#issuecomment-175578758

Does it make the job for you ?

stevezau:
@noirbizarre ah, not sure how i missed the inputs :)

Issue #54, opened 2015-06-25
No way to configure .../swagger.json to go over HTTPS?


Iulian7:
Hello,

I have an application deployed using Flask-RESTPlus. Everything works nicely when it goes over http, but soon as I switch to https, it can't load the swagger.json because it is making requests to http://host/swagger.json and not https://host/swagger.json

The endpoints are registered properly with blueprints, because if I manually go to https://host/swagger.json, I get the correct JSON. However, the generated swagger page tries to load the JSON from a http endpoint, instead of the correct https endpoint. 

Is there any way to fix this?

Thanks

cdumay:
hi, i've got the same problem, i wrote this little monkey patch:

from flask.ext.restplus import Api
from flask import url_for

@property
def specs_url(self):
    """Monkey patch for HTTPS"""
    return url_for(self.endpoint('specs'), _external=True, _scheme='https')

Api.specs_url = specs_url

noirbizarre:
Hi !

I have multiple instances working in both http and https without any problem.

I think this is related to Flask/Werkzeug and the way it handles https.

If you want https, you need to make sure your reverse-proxy is setting the `X-Forwarded-Proto` header properly.
If set, Flask/Werkzeug will set the `request.is_secure` to `True` (this is a good way to check this is correctly detected) and `url_for` will return https URLs (in the same request context).

The solution you provide should be working, but you won't be able anymore to work in http.

noirbizarre:
After some research, you need to apply these:
http://flask.pocoo.org/docs/0.10/deploying/wsgi-standalone/#proxy-setups

The ``ProxyFix`` is what you need:
```python
from werkzeug.contrib.fixers import ProxyFix
app.wsgi_app = ProxyFix(app.wsgi_app)
```

I works well (I don't know why it's not the default).

I modified the todo example to include it.

frol:
@noirbizarre 

> I works well (I don't know why it's not the default).

There is a warning there explaining why it is not the default:

> Please keep in mind that it is a security issue to use such a middleware in a non-proxy setup because it will blindly trust the incoming headers which might be forged by malicious clients.

d-kahara:
For anyone that's encountered this issue lately
This has **changed** 
>from werkzeug.contrib.fixers import ProxyFix
app.wsgi_app = ProxyFix(app.wsgi_app)

**This is what you need:**
>from werkzeug.middleware.proxy_fix import ProxyFix
[...]

PR #128, opened 2016-01-28
Use relative swagger.json URL for SwaggerUI

drewdogg:
I run my flask-restplus app in the default werkzeug HTTP webserver behind a reverse HTTPS proxy.  The absolute URL of the swagger.json is problematic for the SwaggerUI AJAX call in this scenario.  All of the other resources in the HTML use relative URLs anyways.

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -424,6 +424,15 @@ class Api(object):
         return url_for(self.endpoint('specs'), _external=True)
 
     @property
+    def specs_path(self):
+        '''
+        The Swagger specifications path
+
+        :rtype: str
+        '''
+        return url_for(self.endpoint('specs'))
+
+    @property
     def base_url(self):
         '''
         The API base absolute url
--- a/flask_restplus/apidoc.py
+++ b/flask_restplus/apidoc.py
@@ -33,4 +33,4 @@ def swagger_static(filename):
 def ui_for(api):
     '''Render a SwaggerUI for a given API'''
     return render_template('swagger-ui.html', title=api.title,
-                           specs_url=api.specs_url)
+                           specs_url=api.specs_path)


noirbizarre:
Why is that ? Can you precise your initial issue ?
Have you set SERVER_URL in your configuration and added ProxyFix ?
I'm using this setup in production and it's working very well.
See http://flask.pocoo.org/docs/0.10/deploying/wsgi-standalone/#deploying-proxy-setups (this Flask related, not specific to flask-restplus)

I'm not sure that using the path is compatible with subdomain blueprints.

drewdogg:
The issue is the conflict with http and https in the full URL.  flask-restplus generates the URL with http:// but my reverse proxy load balancer only accepts https://, so the swagger.json will never load via AJAX call.  I do not have ProxyFix.  I do not have control over the headers that can be set by the load balancer.

I'm not sure what you mean by subdomain blueprints, but isn't the hostname of swagger.json and the SwaggerUI elements always the same?  Alternatively, we could have the URL with the hostname but without the scheme (i.e. "//server.com/swagger.json") and it will load properly.  Why should flask-restplus should care what the hostname is?

noirbizarre:
No, you can have different URL for documentation and API so absolute URL is necessary.
Blueprint can be registered with an url prefix and/or a given subdomain (see #103).

Your reverse proxy (NGinx or HAProxy or whatever) send an `X-Forwarded-Proto` header which is standard which and handled by ProxyFix (it's not a new dependency but a Werkzeug middleware, see the previous link).
This is how Flask is meant to run (and which is explain in the link).
See also #54 because someone already asked for this.

Just so you know, I have the same setup in production (a frontal load balancer performing the SSL offload itself and talking with my app in HTTP) and it's been working fine with ProxyFix for almost 2 years.

drewdogg:
Ah OK.  Would dropping the scheme be a compromise?  Its generally a best practice in web sites to not require http/https scheme in AJAX URLs in HTML, so that it picks up the parent page's scheme.

harishkashyap:
@noirbizarre could we get an nginx and other sample configuration files uploaded with the examples as this is a common scenario.

kuldeeprishi:
Hi @noirbizarre how can i change the swagger.json url. since i don't want to serve swagger.json on hostname/swagger.json. Instead i want to host it at different url such as hostname/_svc/api/swagger.json

plutov:
Can it be merged? I am having issue running Flask on HTTPS.

ajmeese7:
Exact same problem here, this is a huge blocker for me because HTTPS won't work at all behind my reverse proxy

PR #135, opened 2016-02-14
Use OrderedDict for Models

awiddersheim:
Had to do some wierd stuff with `args` to get things to work with `OrderedDict`. Otherwise the `deepcopy()` in `resolved` would not work properly. Struggled finding a better way to do this but would love to hear any ideas.


awiddersheim:
Hrm... clearly my method doesn't work on every Python version.

awiddersheim:
Went another route. @noirbizarre let me know what you think.

PR #136, opened 2016-02-18
Replaced hardcoded list of HTTP methods with `flask.views.http_method_funcs`

frol:
Please, notice that `http_method_funcs` include `TRACE` HTTP method which was not on the list previously, but it seems to be not a big deal.


noirbizarre:
Thanks, it's way better !
Supporting trace should not be a big deal, it may be useful to some using it :+1:

PR #144, opened 2016-03-04
Fix bundled errors

awiddersheim:
Things passed to `handle_validation_error()` must be of type
`ValueError()` for things to work properly.


fizbin:
I wasn't aware of your pull request when I wrote https://github.com/noirbizarre/flask-restplus/pull/175 - we should merge the two efforts.

awiddersheim:
@fizbin Sure! Did you have something specific in mind? My PR was mainly to fix issues that cropped up after @noirbizarre removed the dependency on Flask-RESTful and is basically a reversion of some things back to how they were previously.

PR #150, opened 2016-03-08
Add utf-8 support and make it the default

awiddersheim:


noirbizarre:
Sorry, this one was merged a little bit too fast and I had to revert it.
`application/json` does not accept charset parameter and is UTF-8 by default.
See:
- https://github.com/pallets/flask/issues/454
- http://stackoverflow.com/questions/13096259/can-charset-parameter-be-used-with-application-json-content-type-in-http-1-1

PR #155, opened 2016-03-31
Updated API to make it work under an HTTPS load balancer.

oroszgy:
The previous solution hardcoded the absolute url in the HTML file, which caused problems, when a web service is served behind HTTPS proxy.

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -417,11 +417,11 @@ class Api(object):
     @property
     def specs_url(self):
         '''
-        The Swagger specifications absolute url (ie. `swagger.json`)
+        The Swagger specifications relative url (ie. `swagger.json`)
 
         :rtype: str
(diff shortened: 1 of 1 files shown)


noirbizarre:
I have it running behind an HTTPS load balancer.
Not using `_external=True` here breaks usage with subdomain blueprints.

I think what you need to have correct full URLs resolution is to include ProxyFix middleware and ensure the right headers are set by your load balancer (HAProxy, NGinx... they all handle this).
See http://flask.pocoo.org/docs/0.10/deploying/wsgi-standalone/#deploying-proxy-setups

let me know if it helps and fix you issue.

adolli:
https://github.com/noirbizarre/flask-restplus/pull/599?w=1 
I also create one, but I thought the author is too busy to have this reviewed.

PR #157, opened 2016-04-15
model definitions should report their type

fixedd:
Definitions not having types makes it so that clients such as bravado can't figure out how to unmarshall a response. This should fix that.

I'm not certain that it's required by the standard, but every example I could find contained it and we know that at least one client expects it.


fixedd:
I'm not certain that it's required by the standard, but every example I could find contained it and at least one client expects it.

PR #162, opened 2016-04-21
updated and integrated api and swagger for tags on paths and vendor tags

kmalyar-tic:
Hi!

I just got approval.  

**3 changes that I made was to allow for**

1. custom tags to be added at the path level
2. 3rd party vendor fields
3. having an explicit hostname added to the document (Example is if the doc or swagger ui should be communicating with a host and not running the service local to the doc)

This is personally my first attempt at a pull request, so please, be gentle.

Regards,
Kelvin


kmalyar-tic:
Hi,

To fix these issues, is there an easy way of doing this, or keep resubmitting them? 

I'm not sure why it broke as nosetests all passed locally (and the 2.x tests seem to pass in the Jenkins env).

Issue #132, opened 2016-02-03
hosting behind nginx proxy error


harishkashyap:
I have the Api docs working perfectly locally but when I host with nginx I get this error:
Can't read from server. It may not have the appropriate access-control-origin settings.

I have enabled CORS on /doc/ in nginx. Any ideas? Is there a way to set the url? For prod environment, @noirbizarre is there a way to specify the url? It seems like behind nginx, the swagger.json url doesn't get set properly which can prevent hosting this in prod.

mtrofimm:
+1

ziirish:
I think your answer is [here](http://flask.pocoo.org/docs/0.10/deploying/wsgi-standalone/#proxy-setups).
It is a Flask related thing, not a flask-restplus one.

noirbizarre:
Thanks @ziirish for the response.

@harishkashyap and @mtrofimm, have you tried the ProxyFix solution ? Is it working ?

heaje:
I realize this is closed, but here's another way to fix this same problem via code.

```python
from flask_restplus import Api as rpapi

@property
def specs_url(self):
    """Fixes issue where swagger-ui makes a call to swagger.json over HTTP.
       This can ONLY be used on servers that actually use HTTPS.  On servers that use HTTP,
       this code should not be used at all.
    """
    return flask.url_for(self.endpoint('specs'), _external=True, _scheme='https')
```

Then in your application factory, add something like this (I use a config value called "FORCE_SWAGGER_JSON_HTTPS" to indicate when I want this:
```python
if app.config['FORCE_SWAGGER_JSON_HTTPS']:
    rpapi.specs_url = specs_url
```

webee:
[ProxyFix](http://flask.pocoo.org/docs/0.10/deploying/wsgi-standalone/#proxy-setups)
ensure nginx config set `X-Forwarded-Proto` header
```
{
    ...
    proxy_set_header X-Forwarded-Proto $scheme;
    ...
}
```

Jdubedition:
> I think your answer is [here](http://flask.pocoo.org/docs/0.10/deploying/wsgi-standalone/#proxy-setups).
> It is a Flask related thing, not a flask-restplus one.

If using gunicorn, you need to set the following in a config file:
`secure_scheme_headers = {'X-FORWARDED-PROTOCOL': 'ssl', 'X-FORWARDED-PROTO': 'https', 'X-FORWARDED-SSL': 'on'}`

[...]

PR #163, opened 2016-04-21
swagger vendor tags, host, path tags

kmalyar-tic:
Hi!

I just got approval for this internally.  I've taken the latest pull and run with that:

........................................................................
.............................................................................
## ....

851 tests run in 2.8 seconds (851 tests passed)

3 changes that I made was to allow for

custom tags to be added at the path level
3rd party vendor fields
having an explicit hostname added to the document (Example is if the doc or swagger ui should be communicating with a host and not running the service local to the doc)

I ran nosetests and they all executed 

This is my first attempt at a pull request, so please, be gentle.

Regards,
Kelvin

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -65,6 +65,7 @@ class Api(object):
     :param str contact: A contact email for the API (used in Swagger documentation)
     :param str license: The license associated to the API (used in Swagger documentation)
     :param str license_url: The license page URL (used in Swagger documentation)
+    :param str host: The host that the api is running on.  Not always the localhost, thought it is the default.
     :param str endpoint: The API base endpoint (default to 'api).
     :param str default: The default namespace base name (default to 'default')
     :param str default_label: The default namespace label (used in Swagger documentation)
@@ -82,7 +83,7 @@ class Api(object):
     '''
 
     def __init__(self, app=None, version='1.0', title=None, description=None,
-            terms_url=None, license=None, license_url=None,
+            terms_url=None, license=None, license_url=None, host=None,
             contact=None, contact_url=None, contact_email=None,
             authorizations=None, security=None, doc='/', default_id=default_id,
             default='default', default_label='Default namespace', validate=None,
@@ -99,6 +100,7 @@ class Api(object):
         self.contact_url = contact_url
         self.license = license
         self.license_url = license_url
+        self.host=host
         self.authorizations = authorizations
         self.security = security
         self.default_id = default_id
@@ -156,7 +158,7 @@ class Api(object):
         :param str contact: A contact email for the API (used in Swagger documentation)
         :param str license: The license associated to the API (used in Swagger documentation)
         :param str license_url: The license page URL (used in Swagger documentation)
-
+        :param str host: The host that the api is running on.  Not always the localhost, thought it is the default.
         '''
         self.title = kwargs.get('title', self.title)
         self.description = kwargs.get('description', self.description)
@@ -166,6 +168,7 @@ class Api(object):
         self.contact_email = kwargs.get('contact_email', self.contact_email)
         self.license = kwargs.get('license', self.license)
         self.license_url = kwargs.get('license_url', self.license_url)
+        self.host = kwargs.get('host', self.host)
         self._add_specs = kwargs.get('add_specs', True)
 
         # If app is a blueprint, defer the initialization
--- a/flask_restplus/swagger.py
+++ b/flask_restplus/swagger.py
@@ -178,7 +178,7 @@ class Swagger(object):
             'tags': tags,
             'definitions': self.serialize_definitions() or None,
             'responses': responses or None,
-            'host': self.get_host(),
+            'host': self.api.host or self.get_host(),
         }
         return not_none(specs)
 
@@ -305,6 +305,7 @@ class Swagger(object):
                 continue
             path[method] = self.serialize_operation(doc, method)
             path[method]['tags'] = [ns.name]
+            path[method]['tags'].extend(doc[method].get('tags', []))
         return not_none(path)
 
     def serialize_operation(self, doc, method):
@@ -325,8 +326,18 @@ class Swagger(object):
                 operation['consumes'] = ['multipart/form-data']
             else:
                 operation['consumes'] = ['application/x-www-form-urlencoded', 'multipart/form-data']
+        operation.update(self.vendor_fields(doc, method))       
         return not_none(operation)
-
+    
+    def vendor_fields(self, doc, method):
+        '''Extract custom 3rd party Vendor fields prefixed with x-
+           https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#specification-extensions
+        '''
+        exts = {}
+        for key, value in doc[method].iteritems():
+            if key.startswith('x_'):
+                exts.update({key.replace('x_', 'x-'): value})
+        return exts
     def description_for(self, doc, method):
         '''Extract the description metadata and fallback on the whole docstring'''
         parts = []


kmalyar-tic:
I don't have a python 3 environment.  I"m trying to get one to get it to build again.  I have no idea why parts that I didn't have anything to do with are breaking.

noirbizarre:
Thanks !
Don't worry, let's iterate together on this PR, I'm really glad you had approval for it.

I'll start reviewing today (after I released the 0.9.1).

From what I see, you'll be needing:
- some tests.
- some documentation about it.

About Python 3, what OS are you using ?
- Ubuntu: use the deadsnakes PPA: https://launchpad.net/~fkrull/+archive/ubuntu/deadsnakes
- OSX: Homebrew allow Python3 and others version too
- Archlinux: all versions are installable from AUR

Once they are installed, use `tox` to run tests on the different version: https://github.com/noirbizarre/flask-restplus/blob/master/CONTRIBUTING.rst

flask_restplus/api.py
```diff
@@ -65,6 +65,7 @@ class Api(object):
     :param str contact: A contact email for the API (used in Swagger documentation)
     :param str license: The license associated to the API (used in Swagger documentation)
     :param str license_url: The license page URL (used in Swagger documentation)
+    :param str host: The host that the api is running on.  Not always the localhost, thought it is the default.
```
noirbizarre: There is many closed issue on the host subject (ex: #132, #155, #128, #54).
You should have to enter it manually as Flask already handle this properly. If you have to specify the host elsewhere from the `SERVER_NAME` setting, it's propably that you're lacking the [ProxyFix](http://flask.pocoo.org/docs/0.10/deploying/wsgi-standalone/#proxy-setups)

I should had a dedicated section in the documentation.

flask_restplus/swagger.py
```diff
@@ -325,8 +326,18 @@ def serialize_operation(self, doc, method):
...
-
+    
+    def vendor_fields(self, doc, method):
+        '''Extract custom 3rd party Vendor fields prefixed with x-
+           https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#specification-extensions
+        '''
+        exts = {}
+        for key, value in doc[method].iteritems():
+            if key.startswith('x_'):
+                exts.update({key.replace('x_', 'x-'): value})
```
noirbizarre: For a single value, `exts[key.replace('x_', 'x-')] = value` should be sufficient, no ?

kmalyar-tic: Definitely does, I made this change.

flask_restplus/api.py
kmalyar-tic: I am still a bit confused by this.  I have added: 

```
from werkzeug.contrib.fixers import ProxyFix
app.wsgi_app = ProxyFix(app.wsgi_app)
```
to my code and it does not return a host in the swagger doc.
{
basePath: "/v1",
consumes: [
"application/json"
],
info: {
description: "API  Interface",
title: "Tudor Instrument API",
version: "1.0.0"
},
paths: {},
produces: [],
responses: {},
swagger: "2.0",
tags: []
}


When setting SERVER_NAME explicitly it causes errors. (404)

Could you elaborate on how that should work.

Also,  I want to re-confirm the use case.

We are generating a swagger document.  Storing in a mongo db.  Calling it from another location for dynamic function generation.

It looks to me like if I am setting SERVER_NAME it impacts the routing.  I was hoping for a solution where the app behavior isn't driven by the doc.

That said, I don't have a huge amount of experience in this area. Could you provide a bit more clarification on:
  
1) how the proxy fix should behave
2) whether I can have a completely arbitrary host name

kmalyar-tic:
In response to the higher level question.

I am using Rhel 6.5 (can be 7)

I got most of the above working.  I have one issue unfortunately with the tests.

Nosetests pass while the tox tests fail.  I'm not sure of why/where that is.

```
__________________________________________________________________________________________ summary ___________________________________________________________________________________________
ERROR:   py26: InterpreterNotFound: python2.6
  py27: commands succeeded
ERROR:   py33: InterpreterNotFound: python3.3
ERROR:   py34: commands failed
ERROR:   py35: InterpreterNotFound: python3.5
ERROR:   pypy: InterpreterNotFound: pypy
ERROR:   pypy3: InterpreterNotFound: pypy3
  doc: commands succeeded
(restplus)  ~/git/flask-restplus [master] $ nosetests
.............................................................................
.............................................................................
.............................................................................
.............................................................................
.............................................................................
.............................................................................
.............................................................................
.............................................................................
.............................................................................
.............................................................................
.............................................................................
...................................................

-----------------------------------------------------------------------------
898 tests run in 1.797 seconds (898 tests passed)
(restplus)  ~/git/flask-restplus [master] $ python
Python 3.4.3 (default, Jan 26 2016, 02:25:35)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-4)] on linux
```

I'm still trying to figureout how to identify the bug.  with the python 3 code.

bedge:
This may be a religious issue here, but is it possible to disable a feature for python 3 if it's only functional in 2?
Seems like this effort died a few months ago and if there's no integration soon the work will have been for naught.

kmalyar-tic:
@bedge 
If that is all it is, I would be happy to just provide documentation and unit testing .  I am just not sure what to do with py3 at the moment.  I not really able to give it the dedicated attention I need to.

bedge:
@kmalyar-tic I rebased your work to master with python 3 disabled, just to see if everything else still passed.
https://github.com/noirbizarre/flask-restplus/pull/191, from git@github.com:bedge/flask-restplus.git
If you have some docs & unit tests to commit, I may have some time to poke at the python 3 tests. I have a local config where I can at least verify the same failures.

bedge:
@kmalyar-tic I took a pass at the python3 stuff and just merged in some 2to3 recommendations and it fixed the failing pythin3 unit tests.
Now the only holdup is the unit test coverage drop.
If you have anything you can add on the coverage side, this should be good. feel free to pyll sown my PR and re-submit with your work if you want, https://github.com/noirbizarre/flask-restplus/pull/192

noirbizarre:
Sorry for the late response.

I think you should rebase on latest master to make it pass.

Great for the tag part, but for the host part, I think it should be in because [Flask already provide all necessary to proprely specify host](http://flask.pocoo.org/docs/0.11/deploying/wsgi-standalone/#deploying-proxy-setups). Losts of Flask internals relies on it and I don't want to monkeypatch or bypass this part as it leads to lots of inconsistency.

kmalyar-tic:
Will do my best to get this working this weekend :)  Thanks!

I understand a desire not to monkey patch.  With our use case, we are generating the swagger docs, but they they are being collected elsewhere.  Not only that, but they are also aliased due to load balancers.  I know that Flask provides some information on how to properly specify host, but trying that, it did not then work running the service.  

Can you help me out with an example of how that should happen?  I'm relatively new.

The approach I took was to just add that one tag to make it config driven more than fiddling with either nginx configs or some environment variables.

kmalyar-tic:
@bedge  Thanks again!  Thats awesome, I appreciate it

kmalyar:
Hi @noirbizarre ,

I think I've finally wrapped my head around our disconnect.  I can appreciate the proxy fix solution you've provided.  I've gotten it sorta working on my side, but it doesn't actually fit my needs.

Here's the difference:

We are using the library to generate swagger docs and then publish to an api hub.  The host itself doesn't know what its CNAME is.  We gather these these docs and use them for code generation in other places to make web requests through the load balancer.

I don't have any control over the load balancer, we can't touch it.

Proxying the request is fine, but I need to have a config driven url for the host.  Any change to the flask configs breaks the service functionality.

Maybe I can persuade you? :)

Either that, or maybe you can help with a better example, the simple ProxyFix doesn't do it.  It doesn't generate the doc correctly.

Either way, if no, no harm done.  I'll withdraw this peice of functionality and recreate a pull request with just vendor and path tags.

Thanks!
Kelvin

kmalyar-tic:
Hi @noirbizarre ,

I think I've finally wrapped my head around our disconnect. I can appreciate the proxy fix solution you've provided. I've gotten it sorta working on my side, but it doesn't actually fit my needs.

Here's the difference:

We are using the library to generate swagger docs and then publish to an api hub. The host itself doesn't know what its CNAME is. We gather these these docs and use them for code generation in other places to make web requests through the load balancer.

I don't have any control over the load balancer, we can't touch it.

Proxying the request is fine, but I need to have a config driven url for the host. Any change to the flask configs breaks the service functionality.

Maybe I can persuade you? :)

Either that, or maybe you can help with a better example, the simple ProxyFix doesn't do it. It doesn't generate the doc correctly.

Either way, if no, no harm done. I'll withdraw this peice of functionality and recreate a pull request with just vendor and path tags.

Thanks!
Kelvin

PR #166, opened 2016-04-26
Show how to use the zoo example

RobertDeRose:
This just creates a zoo.py file that shows how to load the zoo module.

### DIFF
--- a/examples/zoo.py
+++ b/examples/zoo.py
@@ -1,0 +1,10 @@
+#!/usr/bin/env python3
+
+from flask import Flask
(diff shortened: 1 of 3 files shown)


noirbizarre:
Hi !

Thanks for your contribution but there is already an example in [complex.py](https://github.com/noirbizarre/flask-restplus/blob/master/examples/complex.py).
It's bad practice to have a module and a package with the same name (I don't even know if the interpreter is able to see both) this is why it's called complex.py instead.

PR #167, opened 2016-04-26
reqparse was missing swagger type number for python type float

danielhauagge:
reqparse was missing the type `number` in `PY_TYPES` for the python type `float`.


danielhauagge:
Hm, I expanded the unit test to check for the changes I incorporated, is there something else I should do?

PR #175, opened 2016-05-28
Fix error bundling, reqparse exception consistency

fizbin:
Since `reqparser.py` was made to not depend on `flask-restful` back in December, error bundling has been broken such that with error bundling on, `parser.parse_args()` doesn't throw an exception on missing required parameters, but simply replaces the missing parameters with the string value of the validation error message (e.g. "Missing required parameter in the JSON body or the post body")

This fixes that, and introduces the consistent exception handling fix also found in https://github.com/flask-restful/flask-restful/pull/589

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,11 @@ Changelog
 
 .. currentmodule:: flask_restplus
 
+Current
+-------
+
+- Nothing yet
+
 0.9.2 (2016-04-22)
 ------------------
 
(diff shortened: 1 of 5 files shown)


fizbin:
Apparently I was not the only to find this and think to fix it: https://github.com/noirbizarre/flask-restplus/pull/144

fizbin:
Closing/reopening to trigger travis-ci to rebuild.

PR #177, opened 2016-05-31
Calling init_app on blueprint fixed

Dlotan:
I guess it made no sense to call init_app on a blueprint here


noirbizarre:
I guess you're right ;)

Issue #176, opened 2016-05-31
Input payload verification of a custom field


aviaryan:
I have created a custom field called `EmailField`. Also I am validating the input payload with api.expect . 
It looks to me that restplus validates the fields only by their data types. Is there away to run a method (eg validate()) for a custom field at input payload verification.

In resource -> validate_payload, I see the comment "# TODO: handle third party handlers" which suggests that this may not be supported. But I am in urgent need of this feature so asked this anyway. 

I can always create a custom function to validate the payload and call it at the start of each `post` method to do the job.

noirbizarre:
Fixed by #178

PR #178, opened 2016-06-03
Add semantic validation for input payload

tzabal:
The JSON Schema specification defines the optionally-supported 'format' keyword for semantic validation of string values. [[1]](http://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-7)

For example, given the following JSON schema:
```json
{
  "type": "object",
  "properties": {
    "name": {"type": "string", "format": "ipv4"}
  }
}
```
A JSON schema validator that does not conform to the 'format' keyword would render the JSON object `{"ip": "awesome-ip"}` valid, whereas one that conforms to it, would render it invalid.

Currently Flask-RESTPlus does not support the ability to add a semantic validator. With this patch, the Api constructor may receive a `jsonschema.FormatChecker` object as argument, that will be passed from the `resource.Resource.validate_payload` method to the `model.Model.validate` method, and will be used in the invocation of the `Draft4Validator` that actually performs the validation.

Resolves issue #176.

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -15,7 +15,7 @@ from flask import make_response as original_flask_make_response
 from flask.helpers import _endpoint_from_view_func
 from flask.signals import got_request_exception
 
-from jsonschema import RefResolver
+from jsonschema import RefResolver, FormatChecker
 
 from werkzeug import cached_property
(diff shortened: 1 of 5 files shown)


noirbizarre:
Thanks !! It might be very useful for anyone using payload validation !

PR #184, opened 2016-07-13
Fixed bug where url_regex allowed out of range ipv4 addresses and ports

fixedd:
It was allowing invalid IPs (such as 300.300.300.300 ) and port numbers greater than 65535.


noirbizarre:
Nice.
Add some tests to ensure its working and a squash the commits into a single one, and I'll merge it ASAP

fixedd:
Done

noirbizarre:
It doesn't seems to work with `http://300:300:300:300` (cf. [Travis Build error](https://travis-ci.org/noirbizarre/flask-restplus/jobs/161676299#1-1859))

Issue #185, opened 2016-07-19
ImportError: cannot import name 'Namespace'


merritts:
In working through some of the examples provided in the docs, I am running into an issue with importing the Namespace class. Mainly, it does not appear to be importable. The snipped from the interpreter shows the issue.

```
>>> from flask_restplus import Namespace
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'Namespace'
```

It looks like in ```__init__.py```, ```Namespace``` is imported but not included in ```__all__```:

```
from .namespace import Namespace  # noqa

__all__ = (
    '__version__',
    '__description__',
    'Api',
    'Resource',
    'apidoc',
    'marshal',
    'marshal_with',
    'marshal_with_field',
    'Mask',
    'Model',
    'abort',
    'cors',
    'fields',
    'inputs',
    'reqparse',
    'RestError',
    'SpecsError',
    'Swagger',
    'ValidationError',
)
```

PR #186, opened 2016-07-19
Issue #185 - Namespace Import Error

merritts:
Added ```Namespace``` in ```__all__``` to make it directly importable.


jblack14:
As of 12/29/17, v0.10.1, "Namespace" still does not appear to be included in __all__.

PR #198, opened 2016-09-03
Add custom paths for ns in apis

ufian:
Hello!

I very love your namespace objects, but it isn't flexible. I add support of bind custom prefix path in Api.add_namespace method.

Example of namespace file
```python
#mymodule.py

ns = restplus.Namespace('test_ns', description='Test namespace')

@ns.route('/test/', endpoint='test_resource')
class TestResource(restplus.Resource):
    pass
```

And create application
```python
#app.py

import mymodule

app = flask.Flask(__name__)
api = restplus.Api(app)

api.add_namespace(mymodule.ns, '/api_test')
```

In result I bind `mymodule.ns` on prefix path `'/api_test'`. If you try generate url by `url_for('test_resource')`, then you will get url `/api_test/test/`.

### DIFF
--- a/doc/scaling.rst
+++ b/doc/scaling.rst
@@ -110,6 +110,31 @@ The `apis.__init__` module should aggregate them:
     api.add_namespace(nsX)
 
 
+You can define custom url-prefixes for namespaces during registering them in your API.
+You don't have to bind url-prefix while declaration of Namespace object.
+
+.. code-block:: Python
+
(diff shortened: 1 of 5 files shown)


noirbizarre:
Seems good to me. Can you add a little bit of documentation about this ?

ufian:
I added method declaration and example in doc.

PR #213, opened 2016-11-04
Add CSRF Token Support

ci5er:
This adds some minimal belt-and-suspenders necessary to process CSRF tokens properly (as needed to work in a larger system for Ajax calls because of WTForms embedded within Flask-Login, which is embedded in turn within Flask-Security).

### DIFF
--- a/flask_restplus/templates/swagger-ui.html
+++ b/flask_restplus/templates/swagger-ui.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
+    <meta name="csrf-token" content="{{ csrf_token() }}">    
     <title>{{ title }}</title>
     {% include 'swagger-ui-css.html' %}
     {% include 'swagger-ui-libs.html' %}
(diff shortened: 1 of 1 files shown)


warwing:
Any news on this issue? I would be very interested in this fix/enhancement...

ci5er:
@warwing
I never figured out how to deal with the pre-merge auto-check, and have since gone onto other projects, but if you hand-merge (or copy-and-replace) this single file I submitted (here: https://github.com/noirbizarre/flask-restplus/pull/213/commits/c14776dd04a3f69c39c2623779ddac6fae84446a ) it worked with the version I was dealing with at the time...

warwing:
@ci5er 
Thanks for the fast reply - but this means, that I have to fork this project - because I then have local modifications and can't install it from pypi - right?

I just merged you're changes to 0.10.1 - would this help? Or will all changes be merged to master?
I tried it with master - but master changed from swagger 2.2 to swagger 3.4...
...theoretically it should still work: https://swagger.io/docs/swagger-tools/#parameters-37

ci5er:
Re: Merging to 0.10.1 - I suspect it won't make it upstream (and I don't really use pypi, so I don't know if it would make it there through whatever CI/CD system they are probably using, but I'd guess not).

Re: swagger 2.2 to swaggger 3.4. There was a bug in the 2.0-dev version of swagger-ui that I was using for some reason (maybe that was the one that they were using in 0.10.1?), but the patch for the bug had been merged, but not made it into an actual release. That has surely been done for a long-time now. Also, if you don't have a lot of intercepts (and here there was only this one that I added), that bug didn't get up to bother anyone.

I appreciate you picking up the baton on this little snippet of code. At the time I was getting benefit for the project so I thought I should share back. I couldn't figure out what Travis CI was telling me (I don't use github often - although I do use git. In any case the GUI was confusing me!) so I didn't bother. That was sloppy on  my part not to clean that up.

Issue #223, opened 2016-12-30
Swagger UI refers to http://localhost:5000/swagger.json


rndrestore:
We deploy flask-restplus behind a proxying firewall.

Problem here is that the swagger.html generates this:

url: "http://127.0.0.1:5000/swagger.json"

This is fine for local deployments, but blocked when proxied through a firewall.

The problem come from api.py:

@property
    def specs_url(self):
        '''
        The Swagger specifications absolute url (ie. `swagger.json`)

        :rtype: str
        '''
        return url_for(self.endpoint('specs'), _external=True)

Using _external=False results in '/swagger.json' and solves this issue, and nicely serves swagger UI via the proxying firewall.

ziirish:
I think your issue is Flask related. Have a look at #132.

andrecp:
I have the same issue, I've spent the whole day debugging it and hopefully got somewhere, my case:

I have two flask-restplus apps running, one in localhost:5000 and other in localhost:5100. I am using the example on the README.md for both of them. I am deploying them behind nginx.

My intention is that localhost/contacts proxies to localhost:5000 and localhost/notifications proxies to localhost:5100.

With the below changes, things work *almost* as expected:
*  When accessing localhost/contacts/todos/ it's all good;
*  When accessing localhost/contacts/todos I get redirect to localhost/contacts/todos/ thanks to the proxy_redirect line;
* When accessing localhost/contacts it fails to load swagger ui. It tries to read swagger.json from localhost/swagger.json instead of localhost/contacts/swagger.json. I can't for the life of me make it work with nginx changes only.

```
server {

     listen 80;
     server_name localhost;

     location /contacts/ {
         proxy_pass http://localhost:5000/;
         proxy_redirect http://localhost http://localhost/contacts;

         proxy_set_header   Host             $host;
         proxy_set_header   X-Real-IP        $remote_addr;
         proxy_set_header   X-Forwarded-For  $proxy_add_x_forwarded_for;
         proxy_set_header   X-Forwarded-Proto    $scheme;
         proxy_set_header   Referer $http_referer;
     }
}
```

The solution was to follow http://flask.pocoo.org/snippets/35/ and add the following to nginx block:

```
    proxy_set_header   X-Scheme $scheme;
    proxy_set_header   X-Script-Name /contacts/;
```
And then call flask-restplus with

```
class ReverseProxied(object):
    def __init__(self, app):
        self.app = app

    def __call__(self, environ, start_response):
[...]

PR #226, opened 2017-01-07
Suggestion to make swagger work under a reverse proxy

andrecp:
Hey,

Did a small fix trying to solve this issue: https://github.com/noirbizarre/flask-restplus/issues/223 , I've seen it appearing at least 3 times and is always the swagger.json that is the problem. 

It's a little bit of a hack but happy to discuss if the logic makes sense. If it does I can update the docs / add tests / make it properly and then update the PR.

Cheers

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -367,7 +367,7 @@ class Api(object):
 
         Endpoints are ensured not to collide.
 
-        Override this method specify a custom algoryhtm for default endpoint.
+        Override this method specify a custom algorithm for default endpoint.
 
         :param Resource resource: the resource for which we want an endpoint
         :param Namespace namespace: the namespace holding the resource
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -397,7 +397,7 @@ class Api(object):
         '''
         This method registers resources from namespace for current instance of api.
         You can use argument path for definition custom prefix url for namespace.
-        
+
         :param Namespace ns: the namespace
         :param path: registration prefix of namespace
         '''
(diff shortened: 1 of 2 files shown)


mrasband:
Is this fixed if you just use what is suggested in [proxy setups](http://flask.pocoo.org/docs/0.12/deploying/wsgi-standalone/#proxy-setups)?

e.g.

```
from werkzeug.contrib.fixers import ProxyFix
app.wsgi_app = ProxyFix(app.wsgi_app)
```

in the app factory.

This should delegate to the flask `url_for` handling for generating some of the urls, and that respects the `X-Forwarded-*` headers from something like nginx.

andrecp:
Last time I checked no it wasn't. The proxy setup works fine if you're serving the swagger docs from the root but if you change the swagger docs URL behind a reverse proxy it doesn't work.

That said, I've been serving my docs from root URL and haven't had the need for this anymore.

czobrisky:
I know this is an old PR, but I'm currently having this issue behind a reverse proxy and would like to see the issue resolved and am willing to work on it if required.  Are these changes still valid?

cacoyle:
I think I was able to accomplish the same desired effect with a smaller change

```
api = Api(
        app,
        doc="/foobar/api/"
)
```

```
--- flask_restplus/api_orig.py	2018-10-18 15:52:40.869054234 -0400
+++ flask_restplus/api.py	2018-10-18 15:52:26.728870056 -0400
@@ -241,7 +241,7 @@
                 app_or_blueprint,
                 SwaggerView,
                 self.default_namespace,
-                '/swagger.json',
+                self._doc + 'swagger.json',
                 endpoint=endpoint,
                 resource_class_args=(self, )
             )

```

wmolicki:
I'd love to have the above fix merged

PR #237, opened 2017-02-14
Fix typo in __init__.py

mcs07:
Missing comma between 'SchemaModel' and  'abort' was causing error:

     AttributeError: module 'flask_restplus' has no attribute 'SchemaModelabort'

when doing `from flask_restplus import *`

### DIFF
--- a/flask_restplus/__init__.py
+++ b/flask_restplus/__init__.py
@@ -23,7 +23,7 @@ __all__ = (
     'marshal_with_field',
     'Mask',
     'Model',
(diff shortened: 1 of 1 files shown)


frol:
@noirbizarre Could you please take an action on the PRs?

frol:
OK, this was fixed with #242, so this PR should be closed.

@noirbizarre Thank you!

noirbizarre:
Same as #242 
Thanks !

Issue #243, opened 2017-02-23
swagger.json does not support custom URL converters


eddy-geek:
Similar issue to #233.

Minimal working example (to enable parsing the URL according to a regular expression):
```python
from flask import Flask
from werkzeug.routing import BaseConverter
from flask_restplus import Api, Resource


class RegexConverter(BaseConverter):
    def __init__(self, url_map, *items):
        super(RegexConverter, self).__init__(url_map)
        self.regex = items[0]


app = Flask(__name__)
app.url_map.converters['regex'] = RegexConverter
api = Api(app, version='1.0', title='RegexExample')


@api.route('/<regex("[0-9]*"):uid>/')
# @api.route('/<uid>/')
class Example(Resource):
    def get(self, uid):
        return {'success': uid}


if __name__ == '__main__':
    app.run(debug=True)
```

* With `@api.route('/<uid>/')` everything is obviously working:

![image](https://cloud.githubusercontent.com/assets/2772505/23256381/7786129e-f9bf-11e6-9f98-e51efaa6748e.png)

* With `@api.route('/<regex("[0-9]*"):uid>/')` it all explodes:
```
  File "/home/user/.local/lib/python3.5/site-packages/flask_restplus/swagger.py", line 82, in extract_path_params
    raise ValueError('Unsupported type converter: {}, {}'.format(descriptor, name))
ValueError: Unsupported type converter: regex("[0-9]*"), uid
```

noirbizarre:
Fixed by #244 
(Basic support, will probably be enhanced in 0.11)

PR #244, opened 2017-02-23
Add support for optional type arguments in URLs e.g. string(length=2).

### LINKED ISSUE #233: Swagger doesn't support converters with optional arguments
If one defines a route with optional arguments, http://werkzeug.pocoo.org/docs/0.11/routing/#rule-format, i.e., `@api.route('/my-resource/<string(length=2):id>')`, Swagger raises a `ValueError` (swagger.py#L82) because in senses the type converter is unsupported, i.e.,

```python
from flask import Flask
from flask_restplus import Api, Resource


app = Flask(__name__)
api = Api(app)


@api.route('/my-resource/<string(length=2):id>')
class MyResource(Resource):
    def get(self, id):
        return id

    @api.response(403, 'Not Authorized')
    def post(self, id):
        api.abort(403)


if __name__ == '__main__':
    app.run(debug=True)
```

The workaround seems to be to explicitly register the converter via, 

```python
from werkzeug.routing import UnicodeConverter
app.url_map.converters['string(length=2)'] = UnicodeConverter
```

which seems somewhat kludgy, especially when one has multiple signatures which use arguments.

eddy-geek:
Fixes #233 and #243.

### DIFF
--- a/flask_restplus/swagger.py
+++ b/flask_restplus/swagger.py
@@ -68,6 +68,9 @@ def extract_path_params(path):
     params = OrderedDict()
     for match in RE_PARAMS.findall(path):
         descriptor, name = match.split(':') if ':' in match else (None, match)
+        if descriptor and '(' in descriptor:  # optional arg e.g. string(length=2)
+            descriptor = descriptor.split('(')[0]
+
         param = {
             'name': name,
             'in': 'path',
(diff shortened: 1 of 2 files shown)


eddy-geek:
Indeed that seems like the proper way to do it, though it means refactoring a bit of the existing code (`extract_path_params` at least).

Until someone finds the time for that, this PR seems to cover the issues reported and is easy to merge ;-)

noirbizarre:
@johnbodley Thanks for the tips, I'll try to refactor to use this. It might be way safer and powerful.
@eddy-geek Thanks for the PR 👍

PR #248, opened 2017-03-04
Preserve order of parameters

frol:
It is useful for Swagger Codegen since most of the programming languages rely on the function parameters order.

### DIFF
--- a/flask_restplus/swagger.py
+++ b/flask_restplus/swagger.py
@@ -220,7 +220,7 @@ class Swagger(object):
         if doc is False:
             return False
         doc['name'] = resource.__name__
-        params = merge(self.expected_params(doc), doc.get('params', {}))
+        params = merge(self.expected_params(doc), doc.get('params', OrderedDict()))
         params = merge(params, extract_path_params(url))
         doc['params'] = params
         for method in [m.lower() for m in resource.methods or []]:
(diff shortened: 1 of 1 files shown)


noirbizarre:
👍

frol:
@noirbizarre Thank you for the lightning fast response!

noirbizarre:
@frol It compensate a little bit the last year ;)

propertone:
This doesn't completely work. Method specific parameters using the RequestParser are still unsorted. I believe the following patch to swagger.py would fix this:

```diff
@@ -233,7 +233,7 @@
                 method_doc['docstring'] = parse_docstring(method_impl)
                 method_params = self.expected_params(method_doc)
                 method_params = merge(method_params, method_doc.get('params', {}))
-                inherited_params = dict((k, v) for k, v in iteritems(params) if k in method_params)
+                inherited_params = OrderedDict((k, v) for k, v in iteritems(params) if k in method_params)
                 method_doc['params'] = merge(inherited_params, method_params)
             doc[method] = method_doc
         return doc
@@ -245,7 +245,7 @@
 
         for expect in doc.get('expect', []):
             if isinstance(expect, RequestParser):
-                parser_params = dict((p['name'], p) for p in expect.__schema__)
+                parser_params = OrderedDict((p['name'], p) for p in expect.__schema__)
                 params.update(parser_params)
             elif isinstance(expect, ModelBase):
                 params['payload'] = not_none({


```

frol:
@propertone I agree that this might be the case, however, I use my custom patchset with Webargs handling parameters, so I cannot test if your patch is working (but it should as far as I can see). Go ahead and propose a PR!

PR #250, opened 2017-03-06
Provide an HTTP cache decorator

davidbgk:
WIP: to be able to discuss the cache pattern to compare etags.


ziirish:
Hello,

Thanks for your PR and sorry for this **late** answer...
I like the idea, but this definitely needs more work such as making all the tests pass, fixing the coverage decrease and documenting the usecase.

Would you mind giving us some more context about this change?
I can take care of the rest.

davidbgk:
> Would you mind giving us some more context about this change?

It's been a while :)

If I remember correctly it was to be able to set proper cache headers for a given view. This way clever clients can reduce the number of requests (like just checking the headers prior to request the whole body).

I don't use anymore flask-restplus on a daily basis so it will be harder for me to get the motivation to continue on that PR without a clear goal. You can close it if you want.

ziirish:
No problem, I understand it's been a long time since you proposed this improvement.

So you wouldn't mind if we finish this work?

I'll propose this to our roadmap because I like that feature.

davidbgk:
> So you wouldn't mind if we finish this work?

Sure, I would be super happy if it is eventually available somehow 👍

Issue #249, opened 2017-03-05
Swagger documentation incorrectly contains Sphinx docstring fields


johnbodley:
The Swagger implementation notes contains numerous Sphinx docstring fields which I presume should be removed,

<img width="1056" alt="screen shot 2017-03-04 at 8 34 03 pm" src="https://cloud.githubusercontent.com/assets/4567245/23584601/93a034e8-011a-11e7-9292-7f2705d86eca.png">

```python
from flask import Flask
from flask_restplus import Api, Resource


app = Flask(__name__)
api = Api(app)


@api.route('/my-resource/<int:id>')
class MyResource(Resource):
    def get(self, id):
        """
        Some summary.
        
        :param id: The ID
        :type id: int
        :returns: Nothing
        :rtype: None
        :raises SomeException: in case of something
        """
        
        pass


if __name__ == '__main__':
    app.run(debug=True)
```

i.e., per the [parse_docstring](https://github.com/noirbizarre/flask-restplus/blob/85932fbabb9186043d7d03e703ae6f3971fa97ca/flask_restplus/swagger.py#L102) method only the `raises` field is ignored. Note [This](http://stackoverflow.com/a/11315548/1011716) seems like the correct approach for parsing a docstring containing Sphinx formatting.

@noirbizarre I'm happy to try to tackle this issue, though I need some additional context. Which Sphinx fields to you care about? Currently the `parse_docstring` method populates only a subset of the Sphinx fields (and defaults `returns` and `params` to `None` and `[]` respectively). Does Swagger leverage these fields elsewhere or is it suffice to merely provide the `summary` (first paragraph) and `details` (remaining paragraphs)? If that's the case the following logic may work, 

```python
from docutils.core import publish_doctree
import xml.etree.ElementTree as ET

raw = """
Some summary. 

This is another paragraph. This is second sentence in the same paragraph.

This is a third paragraph.

:param id: The ID
:type id: int
:returns: Nothing
:rtype: None
:raises SomeException: in case of something
"""

tree = ET.fromstring(publish_doctree(raw).asdom().toxml())
elems = tree.findall('paragraph')

parsed = {
    'raw': raw, 
    'summary': elems[0].text if elems else None,
    'details': '\n'.join([e.text for e in elems[1:]]) if len(elems) > 1 else None
} 
```

PR #252, opened 2017-03-12
[swagger] Parse docstring

johnbodley:
This PR addresses https://github.com/noirbizarre/flask-restplus/issues/249 which uses `docutils` to parse the docstring and extract the top level paragraphs and fields.

A few notes:

1. The `params` and `returns` keys were removed as they were never set.
2. Per the `docutils` parsing the `summary` and `details` fields are defined via top-level summary paragraphs, i.e., a blank line, rather than a newline as was previously the case.
3. I was unable to successfully test the entire repo even from the master branch using `inv test` per the instructions in the CONTRIBUTING.rst file. I was able to run, 

```
inv qa
pytest tests/test_swagger_utils.py 
```

### DIFF
--- a/flask_restplus/swagger.py
+++ b/flask_restplus/swagger.py
@@ -3,7 +3,9 @@ from __future__ import unicode_literals, absolute_import
 
 import itertools
 import re
+import xml.etree.ElementTree as ET
 
+from docutils.core import publish_doctree
 from inspect import isclass, getdoc
 from collections import OrderedDict, Hashable
 from six import string_types, itervalues, iteritems, iterkeys
@@ -38,8 +40,6 @@ RE_URL = re.compile(r'<(?:[^:<>]+:)?([^<>]+)>')
 
 DEFAULT_RESPONSE_DESCRIPTION = 'Success'
 DEFAULT_RESPONSE = {'description': DEFAULT_RESPONSE_DESCRIPTION}
-
-RE_RAISES = re.compile(r'^:raises\s+(?P<name>[\w\d_]+)\s*:\s*(?P<description>.*)$', re.MULTILINE)
 
 
 def ref(model):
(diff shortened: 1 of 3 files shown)


johnbodley:
@noirbizarre I was wondering whether I could get your input on this PR. It's not clear to me whether the approach I took is what your intention was with regards to parsing of the docstring.

brezniczky:
@john-bodley have you noticed that your commit breaks the tests? (was browsing around by accident while checking if others' errors have similarities to one of mine)

john-bodley:
@brezniczky do you know why the tests are failing? This PR is quite old, but I sense I wasn't able to successfully test the master branch and felt it could have been an issue with my configuration.

Kieran7741:
@SteadBytes, @a-luna, @j5awry, @ziirish. Can this be revisited?

j5awry:
We'll need to look at the original issue. Looking at the implementation , I have some reservations. These aren't small libraries to bring in, and I'm not sure what they're buying us. But we'll need to see what the original issue is, and how we want to address it. With the Pr being 2+ years old, I don't envision this being pulled in.

SteadBytes:
I agree with @j5awry that these bringing in these libraries is not ideal, furthermore this introduces quite a significant amount of work to application startup. I think it may be worth re-considering the docstring parsing behaviour and potentially re-work it - i.e. should it be used only for descriptions - ignoring all parameters e.t.c and assuming docstring are for the developer usage and *not* for documenting the API. Or should the docstrings be treated more 'first class' in terms of generating Swagger specs - if so how should they interact with the other documentation methods `@api.doc` e.t.c.

Kieran7741:
@SteadBytes as a quick fix can a flag be introduce to toggle the use of doc strings as descriptions? The summary field is working perfect, it is just the description field that is being populated with sphinx(param, type, rtype, returns).

PR #258, opened 2017-03-17
Add short descriptions for expected model (fix #61)

### LINKED ISSUE #61: Documenting Input Body
I'm having trouble getting JSON body inputs documented in swagger and working the way I want.

My end goal is to setup a `POST` request that is JSON that looks like `{"type": "foobar"}`.

I want to be able to validate the input coming in so I first tried doing this using the following parser:

```
parser = api.parser()

packages_post = parser.copy()
packages_post.add_argument(
        "data",
        required=True,
        location="json",
        help="The type of package to create."
)
```

The problem with this method is it doesn't really document what the schema looks like very well:

image

I then went with a model and used the `body` option in `api.doc()` which did document things well but doesn't do validation:

```
package = api.model(
        "Package",
        {
                "type": fields.String(
                        description="The type of package to create.",
                        required=True
                )
        }
)
```

image

Alright, so I figure I can use a combination of the two to get both the right documentation and validation:

```
package = api.model(
        "Package",
        {
                "type": fields.String(
                        description="The type of package to create.",
                        required=True
                )
        }
)

parser = api.parser()

packages_post = parser.copy()
packages_post.add_argument(
        "package",
        type=package,
        location="json"
)
```

I noticed this pattern in the tests and it almost works however now the input needs to be `{"package": {"type": "foobar"}}` which doesn't get documented very well.

The only way I can see to get things to work the way I want is to define both a parser and a model with the same exact information which is a bit ugly and redundant.

Is there something I might be missing to make this a bit better? Perhaps something I can look at adding to make this possible?

swiesend:
Introduces the `description` keyword to ns.expect() making the documentation of an expected model more intuitive than providing a `tuple` of `(Model, "short description")` as input model.

The regarding issue #61 is closed yet, but wasn't solved as I expected it to be solved.

### DIFF
--- a/flask_restplus/namespace.py
+++ b/flask_restplus/namespace.py
@@ -192,6 +192,7 @@ class Namespace(object):
 
         :param ModelBase|Parse inputs: An expect model or request parser
         :param bool validate: whether to perform validation or not
+        :param str description: short model description
 
         '''
         expect = []
@@ -200,6 +201,8 @@ class Namespace(object):
             'expect': expect
         }
         for param in inputs:
+            if 'description' in kwargs and isinstance(param, Model):
+                param = (param, kwargs.get('description', ''))
             expect.append(param)
         return self.doc(**params)
 


noirbizarre:
Can you add a test case please (so it ensure it will survive to refactorings)

PR #269, opened 2017-04-02
Add an optional key argument, skip_none, in `marshal_with` and `marshal`

calvin620707:
You can skip those fields which values is `None` instead of marshaling those fields with JSON value, null.
This feature is useful to reduce the size of response when you have a lots of fields which value may be None, but which fields are `None` are unpredictable.

### DIFF
--- a/flask_restplus/marshalling.py
+++ b/flask_restplus/marshalling.py
@@ -10,7 +10,7 @@ from .mask import Mask, apply as apply_mask
 from .utils import unpack
 
 
-def marshal(data, fields, envelope=None, mask=None):
+def marshal(data, fields, envelope=None, skip_none=False, mask=None):
     """Takes raw data (in the form of a dict, list, object) and a dict of
     fields to output and filters the data based on those fields.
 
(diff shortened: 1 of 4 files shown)


flask_restplus/marshalling.py
```diff
@@ -44,14 +50,22 @@ def make(cls):
...
     if isinstance(data, (list, tuple)):
-        out = [marshal(d, fields) for d in data]
+        out = [marshal(d, fields, skip_none=skip_none) for d in data]
         if envelope:
             out = OrderedDict([(envelope, out)])
         return out
 
-    items = ((k, marshal(data, v) if isinstance(v, dict)
-              else make(v).output(k, data))
-             for k, v in fields.items())
```
noirbizarre: The generator syntax is important for performance.
You can nest it with another generator (and factorize the `if skip_none`:
```python
if skip_none:
    items = ((k, v) for k, v in items if v is not None)
```
So it's still a generator and the `skip_none` condition is evaluated once

calvin620707: Thanks for the suggestion! Please help to review if the changes meet the requirement :)

nicbou:
@calvin620707, the tests are missing `test_marshal_nested_skip_none`, and it seems like this test would fail, am I mistaken?

calvin620707:
@nicbou Thanks for pointing this out. I had send another PR #327.

PR #306, opened 2017-07-21
Added 'produces' section support for a specific method

frol:
Using `@Namespace.doc(produces=['application/octet-stream'])` decorator for a method, I can now document the mimetype of specific endpoints.

### DIFF
--- a/flask_restplus/namespace.py
+++ b/flask_restplus/namespace.py
@@ -288,6 +288,10 @@ class Namespace(object):
         header.update(kwargs)
         return self.doc(headers={name: header})
 
+    def produces(self, mimetypes):
+        '''A decorator to specify the MIME types the API can produce'''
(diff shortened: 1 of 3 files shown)


frol:
@noirbizarre Please, help me to locate the right place for the test and it would be helpful if you point out to a similar test, so I do it the right way.

noirbizarre:
I think the right place is in [test_swagger.py](https://github.com/noirbizarre/flask-restplus/blob/master/tests/test_swagger.py)

You can see [an example with the deprecated decorator](https://github.com/noirbizarre/flask-restplus/blob/master/tests/test_swagger.py#L2705-L2738)

By the way having a `@api.produces` decorator shortcut would be awesome ;)

frol:
@noirbizarre All done! Please, review.

Issue #194, opened 2016-08-17
AttributeError: Api does not have __schema__ attribute


984958198:
Python 2.7.5 (default, May 15 2013, 22:43:36) [MSC v.1500 32 bit (Intel)] on win32
Type "copyright", "credits" or "license()" for more information.

Flask (0.11.1)
flask-restplus (0.9.2)
Flask-Script (2.0.5)

[2016-08-17 15:07:57,503] ERROR in app: Exception on /mailang/api/v1/swagger.json [GET]
Traceback (most recent call last):
  File "E:\Python27\lib\site-packages\flask\app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "E:\Python27\lib\site-packages\flask\app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "E:\Python27\lib\site-packages\flask_restplus\api.py", line 310, in wrapper
    resp = resource(_args, *_kwargs)
  File "E:\Python27\lib\site-packages\flask\views.py", line 84, in view
    return self.dispatch_request(_args, *_kwargs)
  File "E:\Python27\lib\site-packages\flask_restplus\resource.py", line 44, in dispatch_request
    resp = meth(_args, *_kwargs)
  File "E:\Python27\lib\site-packages\flask_restplus\api.py", line 751, in get
    return self.api.**schema**
  File "E:\Python27\lib\site-packages\flask_restplus\api.py", line 206, in **getattr**
    raise AttributeError('Api does not have {0} attribute'.format(name))
AttributeError: Api does not have **schema** attribute
127.0.0.1 - - [17/Aug/2016 15:07:57] "GET /mailang/api/v1/swagger.json HTTP/1.1" 500 -

matveevtam:
Hi, @noirbizarre !
I have hit this issue yesterday and have some feedback.
The above exception occurs when I try to use a complex model with `@marshal_with()` decorator using a namespace.
This is how I reproduce it.
I took one of your examples `examples/todo/complex.py` with the module `examples/todo/zoo` and modified the file `examples/todo/zoo/cat.py`, lines 5-12:

```
cat = Model('Cat', {
    'id': fields.String(required=True, description='The cat identifier'),
    'name': fields.String(required=True, description='The cat name'),
    'colors': {
        'head': fields.String(),
        'body': fields.String(),
        'paws': fields.String()
    }
})
api.add_model('Cat', cat)

CATS = [
[...]

PR #341, opened 2017-10-20
Adds fields checking to model constructor

elipavlov:
Adds an improvement to model constructor to avoid my case of issue 194.

All test passed for 2.7 and 3.5.3 versions.
Adds few test for demonstrate how it works;

My Implementation restrict developers from creating model fields, which isn't **fields.Raw** childclass.

My issue described here: 
https://github.com/noirbizarre/flask-restplus/issues/194#issuecomment-338188199


elipavlov:
I fixed static analyzer resentment in the PR.

PR #350, opened 2017-11-10
API error handlers read from namespace's error handlers

avilaton:
Registration of error handlers on a namespace created using 
```
class CustomException(RuntimeError): pass

api = restplus.Api(app)
ns = api.namespace("ExceptionHandler", path="/")

@ns.errorhandler(CustomException)
def handle_custom_exception(error):
    return {'message': str(error), 'test': 'value'}, 400
```
fail to be executed because the factory `api.namespace` adds error handlers immediately to `api.error_handlers`.

I made `api.error_handlers` an internal property and on read I enhance with error handlers from child namespaces. Suggestions are welcome.

Thanks for this great package!

### DIFF
--- a/doc/errors.rst
+++ b/doc/errors.rst
@@ -200,3 +200,16 @@ It also allows for overriding the default error handler when used without parame
     Flask-RESTPlus will return a message in the error response by default.
     If a custom response is required as an error and the message field is not needed,
     it can be disabled by setting ``ERROR_INCLUDE_MESSAGE`` to ``False`` in your application config.
+
+Error handlers can also be registered on namespaces. An error handler registered on a namespace
+will override one registered on the api.
+
+
+.. code-block:: python
+
+    ns = Namespace('cats', description='Cats related operations')
+
+    @ns.errorhandler
+    def specific_namespace_error_handler(error):
+        '''Namespace error handler'''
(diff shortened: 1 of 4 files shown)


noirbizarre:
Hi,
Sorry for the very late response and thanks for the contribution.
I'm willing to merge this but I need you to rebase this branch on master and resolve conflicts to do so.

avilaton:
I updated and I'm getting a failure of an unrelated test. Here is where it blows up:
```

def _parse_interval(value):
    '''
    Do some nasty try/except voodoo to get some sort of datetime
    object(s) out of the string.
    '''
    try:
        return sorted(aniso8601.parse_interval(value))
    except ValueError:
        try:
            return aniso8601.parse_datetime(value), None
        except ValueError:
            return aniso8601.parse_date(value), None
```
because `aniso8601.parse_interval(value)` is raising IndexError for the empty string the test is passing. I think I can just add it to the catched exceptions if that is ok, but not sure why this broke down.

avilaton:
Looks like what broke down was due to the recently released aniso8601==3.0.0 from march 9th 2018.

I added some docs and rebased.

noirbizarre:
Thanks !

I fixed it too on my current master but in another way. I'll reapply mine on top of yours :ok_hand:

PR #384, opened 2018-01-22
Updated Swagger version

fbertola:


aklauza:
Is this being merged anytime soon? We are really interested on new swagger UI. Thank you

flask_restplus/templates/swagger-ui.html
```diff
@@ -1,78 +1,91 @@
...
+
+<div id="swagger-ui"></div>
+
+{% include 'swagger-ui-libs.html' %}
+<script>
+    window.onload = function () {
+
+        // Build a system
+        const ui = SwaggerUIBundle({
+            url: "http://petstore.swagger.io/v2/swagger.json",
```
TotempaaltJ: This should be removed, as it's overwritten two lines below

fbertola: You're right. Removed, thanks for the tip!

LoveBootCaptain:
any news on if this will be merged or not? Thx

PR #391, opened 2018-02-07
Add access to payload on namespace object

Olujedai:


danieldaeschle:
@noirbizarre Merge this please. Why wait?

avilaton:
This decreased coverage @danieldaeschle. Do it the same way it is done for API and add tests.

danieldaeschle:
Ok, but i think rhis repo is no longer maintained.

PR #427, opened 2018-05-03
Prevent illegal null values from creeping in

DHager:
**Note:** This PR needs discussion, it is a speculative attempt and not merge-ready.

I'm trying to fix a problem described [here](/noirbizarre/flask-restplus/issues/179#issuecomment-386176439), but apparently it has its own conflicts with the `allow_null` behavior for `Nested` compound items, whose use-case I don't quite understand.

### DIFF
--- a/flask_restplus/fields.py
+++ b/flask_restplus/fields.py
@@ -109,7 +109,7 @@ class Raw(object):
     __schema_example__ = None
 
     def __init__(self, default=None, attribute=None, title=None, description=None,
-                 required=None, readonly=None, example=None, mask=None, **kwargs):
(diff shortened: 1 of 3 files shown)


DHager:
@noirbizarre Any thoughts on this? I'm not sure it's the best solution, but the fundamental issue of "PUT something that matches the schema, then GET it back as something that fails the schema" seems like a definite problem.

DHager:
@noirbizarre Came across this when cleaning out my repos. Not sure whether the bug still exists.

DHager:
Sounds like [the project is dead and being forked](https://github.com/noirbizarre/flask-restplus/issues/770), so I'll just discard this.

PR #461, opened 2018-06-04
Input boolean

ziirish:
Add support for checkbox validation input + consistent behavior between inputs and fields.


ziirish:
Please note this is kind of a breaking change since fields containing `'false'` pr `'0'` will now be considered `False` where they were considered `True`.

PR #470, opened 2018-06-10
Don't include namespaces without resources to the SWAGGER documentation

### LINKED ISSUE #460: Remove default namespace?
This issue has confused me for several days, and I still cannot find a solution to resolve it. Just wanted to see if somebody here has some bright ideas.
For example, I make a namespace called 'test' for my app, but somehow, the default namespace is shown:
image

To get rid of this redundant default namespace, what I have done is to set the default namespace value when initializing the Api:
```python
api = Api(app, default ='test', default_label='this is namespace test')
```
Unfortunately, this is not an ideal solution because I found there were two duplicate namespaces extracted to swagger file.
image

kandziu:
Closes #460

### DIFF
--- a/doc/swagger.rst
+++ b/doc/swagger.rst
@@ -687,6 +687,10 @@ You can hide some resources or methods from documentation using any of the follo
         def delete(self):
             return {}
 
+.. note::
+
+   Namespace tags without attached resources will be hidden automatically from the documentation.
+
 
 Documenting authorizations
 --------------------------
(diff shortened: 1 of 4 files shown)


kandziu:
@TotempaaltJ, can  we describe this behavior in the http://flask-restplus.readthedocs.io/en/stable/swagger.html#hiding-from-documentation as an **Info** block?

TotempaaltJ:
Seems like an okay place.

kandziu:
@TotempaaltJ, documentation updated.

PR #476, opened 2018-06-15
Fix Namespace decorators

### LINKED ISSUE #475: Namespace decorators are not applied to Resources
Reopened #375 because I have a similar issue.

> I have a decorator designed to wrap arbitrary Flask view functions, and I want to apply it to every resource in an Api or Namespace.
> 
> The decorators kwarg in Api.__init__ works as expected. All methods of all Resource objects are decorated and return correct output.
> 
> However, using the analogous kwarg in Namespace has no effect. I've tried every way to wire namespaces to apis to apps in the documentation, and none of them work. In all the wiring strategies, if I use the decorators kwarg on the outer Api, everything works. However, sometimes I may want to apply the decorator only to a sub-tree of the Api, which makes a Namespace-level decorator look like the obvious solution.
> 
> I dug through the source code, and I can clearly see where Api decorators get applied to the Resource methods. However, I don't see where the Namespace decorators factor into the picture. There appear to be no uses of the decorators instance attribute outside of Namespace.init

kandziu:
Closes #475

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ Changelog
 Current
 -------
 
-- Nothing yet
+- Fix Namespace decorators (:issue:`475`)
 
 0.11.0 (2018-05-16)
 -------------------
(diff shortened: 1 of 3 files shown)


avilaton:
Looks legit.

PR #479, opened 2018-06-20
Todo model's id readonly property typo fix

Kamforka:
The `readonly` property had a typo, corrected that, so now the swagger UI displays the example API payload correctly (without the id field), as originally intended.


Kamforka:
Will you merge it?

Issue #513, opened 2018-08-23
Add a flag to turn Swagger-UI 'Try it Out' on/off


AnthonBerg:
Hi!, Thanks for flask-restplus!

I need to be able to turn off the "Try it Out" button in swagger-ui. The method defined here works for us: https://github.com/swagger-api/swagger-ui/issues/3725#issuecomment-334899276 

I've been able to get this going as a patch to `templates/swagger-ui.html` after flask-restplus installation. I've also added it as a flask-restplus config flag which reads `True`/`False` from `config.SWAGGER_UI_TRY_IT_OUT`. PR incoming with an attempt to contribute.

btv:
I would love to see this merged. I'm willing to help, though I don't know how.

parithiban:
@noirbizarre  May i know when will this be merged. Or is there any other way so that try it out option is been disabled for at least `POST` request. 
Something like this would also be useful 

`SUPPORTED_SUBMIT_METHODS` config option  in `[https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md#network] (https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md#network)`

trevorgetty:
I would love to get this feature please.  It is considered a security risk in some production environments and something we do not wish to show by default.  Can we please have a simple config option to disable the TRY IT OUT button for any generated docs thanks.

dosomder:
This is an important feature for production

SteadBytes:
I have added this to our short term roadmap and will look into it shortly - I agree this would be a very good feature :+1:

Colin-b:
How can it be a security risk? Security by obfuscation is not security, it's the opposite.

Laziness will win the fight over properly secured API endpoints. If your endpoint shouldn't be called by someone then add authentication and authorization.

Otherwise one inexperienced Dev may think he 'secured' it while users can still call the endpoint (you'll find more and more tools allowing to easily call endpoints for non developers, not even talking about how obvious it is to bypass this for a Dev)

Introducing this is a security risk, not the other way around :-/

AnthonBerg:
This is simply orthogonal to security. Not really related. In our case we just needed to disable Try it Out because it didn’t work in our setup.

SteadBytes:
Closed #639

PR #514, opened 2018-08-23
#513: add flag to turn Swagger-UI's 'Try it Out' on and off

AnthonBerg:
* add config.SWAGGER_UI_TRY_IT_OUT
* inject DisableTryItOut plugin in swagger-ui.html
* add check for config.SWAGGER_UI_TRY_IT_OUT in swagger-ui.html template which enables DisableTryItOut
* add tests and documentation

The default is that 'Try it Out' is enabled. If set to False then it's disabled.

First Python PR ever! Thanks for your patience :) And thanks for flask-restplus (and swagger-ui)!

### DIFF
--- a/doc/swagger.rst
+++ b/doc/swagger.rst
@@ -928,6 +928,19 @@ you can register a custom view function with the :meth:`~Api.documentation` deco
         return apidoc.ui_for(api)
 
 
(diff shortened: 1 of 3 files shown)


AnthonBerg:
Pardon me, there seems to be a regex issue in 2.7 and pypy.

AnthonBerg:
¯\_(ツ)_/¯

PR #527, opened 2018-09-23
Propagate exceptions from the flask-restplus handler if propagate exceptions is enabled

orishoshan:
Propagate exceptions from the flask-restplus handler even if they come from a restplus resource (for example, if you decorate a restplus resource with Flask-JWT's @jwt_required, then restplus will catch it even if it shouldn't)

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -592,6 +592,13 @@ class Api(object):
         '''
         got_request_exception.send(current_app._get_current_object(), exception=e)
 
+        if not isinstance(e, HTTPException) and current_app.propagate_exceptions:
+            exc_type, exc_value, tb = sys.exc_info()
+            if exc_value is e:
+                raise
+            else:
+                raise e
+
         include_message_in_response = current_app.config.get("ERROR_INCLUDE_MESSAGE", True)
         default_data = {}
 
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -291,12 +291,13 @@ class ErrorsTest(object):
 
         app.config['PROPAGATE_EXCEPTIONS'] = True
 
-        response = client.get('/api/test/')
-        assert response.status_code == 500
-        assert response.content_type == 'application/json'
+        # From the Flask docs:
+        # PROPAGATE_EXCEPTIONS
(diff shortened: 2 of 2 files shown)


orishoshan:
Hey @noirbizarre , is there any chance this can be merged? I've been working with a fork for a while now and would much rather move to working with a release :(

SteadBytes:
@orishoshan Thank you for contributing this! I apologise for lack of activity - I will merge this PR for you now :+1:

cpaulik:
This has broken a few tests for me. If possible adding this to the release notes or the changelog in the documentation would be great.

avilaton:
@SteadBytes @noirbizarre I know this is a very old issue but if you could explain in very few words why this was necessary I would appreciate it very much. I bumped our stuff from 0.12.1 to 0.13.0 and lots of our tests started failing. Could it be that we need to set PROPAGATE EXCEPTIONS to false?
Maybe we need something extra in our app exception handler?
Thanks

SteadBytes:
@avilaton quick disclaimer first:
flask-restplus is no longer maintained. The current maintainers have not been able to produce releases due to lack of pypi privileges. As such, we have created an official fork, flask-restx. On github at python-restx/flask-restx. It's a drop in replacement and has a fix.

Do not expect any releases from flask-restplus in the near future.

Now, without further details of what is breaking it's difficult to diagnose, however it's likely that some third party exceptions (i.e. not from flask-restplus) that were previously being caught are now being propagated up out of the flask-restplus error handlers. I would suggest that you catch these errors and handle them appropriately with additional error handlers 😊

avilaton:
Thanks @SteadBytes, I know about the situation and we are the migration to restx. I found it confusing that new version tags were introduced in the fork. Also, drop in replacement needs to be qualified with a version of restplus, since it was not the case for versions earlier than 0.13.0 which we had.

I found the root cause to this issue is a change in `handle_error` from 0.12.1 to 0.13.0 which appears to be fixed in restx right here https://github.com/python-restx/flask-restx/blob/master/flask_restx/api.py#L657 where it is checked that the exception has no registered handlers before propagating it.

Issue #530, opened 2018-10-05
Custom error handler is discarded if you have a validation error from @api.expect


alexwlchan:
If you have a custom error handler defined, it gets lost when there's a validation error thrown inside `@api.expect()` on a route.

Consider the following example program:

```python
#!/usr/bin/env python
# -*- encoding: utf-8

from flask import Flask
from flask_restplus import Api, Resource, fields

app = Flask(__name__)
api = Api(app)


animal = api.model("Animal", {"name": fields.String(required=True)})


@api.route("/")
class AnimalCollection(Resource):
    @api.expect(animal, validate=True)
    def post(self):
        return '{"status": "ok"}'


@api.errorhandler(Exception)
def default_error_handler(error):
    del error.data
    return {"my_message": "My custom error response (%r)" % error}, 500


if __name__ == "__main__":
    app.run(debug=True)
```

This creates a single endpoint `/` which accepts POST requests. The data sent to this endpoint should be a JSON dictionary of the form `{"name": "..."}`. For example:

```console
$ curl -X POST http://localhost:5000/ -d '{"name": "alex"}' --header 'Content-Type: application/json'
"{\"status\": \"ok\"}"
```

Using the instructions for [error handling in the docs](https://flask-restplus.readthedocs.io/en/stable/errors.html), I've created a custom error handler that adds a `my_message` field to all error responses.

For example, if I were to pass a payload which doesn't validate as an instance of the "animal" model, I'd expect an error like this:

```console
$ curl -X POST http://localhost:5000/ -d '{"legs": 4}' --header 'Content-Type: application/json'
{
    "my_message": "My custom error response (<BadRequest '400: Bad Request'>)",
    "message": "400 Bad Request: The browser (or proxy) sent a request that this server could not understand."
}
```

Instead, if you run the code above, this is the output:

```console
$ curl -X POST http://localhost:5000/ -d '{"legs": 4}' --header 'Content-Type: application/json'
{
    "errors": {
        "name": "'name' is a required property"
    },
    "message": "Input payload validation failed"
}
```

It appears to be bypassing the custom error handler, and presenting the default error that comes from Flask-RESTPlus.

We have a custom Error model for our API that we want to use for all our responses, which is set through an error handler like this – so having Flask-RESTPlus interfering is unhelpful!

### Workaround/diagnosis

Digging around led me to this part of the codebase. I can see why the issue occurs, but not what the resolution is (or what the implications might be):

[...]

PR #552, opened 2018-11-07
Updated handle_error

Yolley:
Updated handle_error, so that custom messages added with errorhandler decorator would be chosen over 'data' attribute in exception, or problems like this https://github.com/noirbizarre/flask-restplus/issues/530 may occur.


hcamino:
Is this going to be included in any future release? When? We are waiting for this to come

Thanks in advance

PR #465, opened 2018-06-04
Adds add_specs to Api

thejamesmarq:
Implements https://github.com/noirbizarre/flask-restplus/issues/464

Issue #464, opened 2018-06-04
Disable swagger.json


thejamesmarq:
In production I'd like to be able to hide the documentation and specs for an API.

I am currently able to prevent documentation from rendering by creating my API with

`Api(app, doc=False)`

but this still allows someone to navigate to /swagger.json and retrieve the file.

It might be worthwhile being able to disable exposing the swagger spec from the app with an option like

`Api(app, doc=False, specs=False`

which would make /swagger.json come back with 404 as expected.

fxrhxn:
Did you ever find a way to make this work? Trying to disable swagger.json as well.

Colin-b:
Hi,

I'm curious about why you'd want to hide the specification of your API. Could you elaborate?

Thanks

myselfhimself:
For enterprise world-available applications. Securing a REST API may
consist in encrypting its connection (HTTPS), adding JWT/Oauth2 etc, and
also hiding the swagger.json in order to prevent any hacker or crawler to
discover the entry points...

Colin-b:
Once your endpoints are using authentication+authorization as you said it should be ok imho.

I wouldn't consider security by obfuscation as an effective additional security but I get that some people do.

Thx for the reply.

noirbizarre:
I agree with @Colin-b, I don't consider this as effective security. But I understand the need and I'm just waiting for the test cases to merge #465.

In the meantime, I'm sure you are not serving flask applications without a reverse-proxy and if so, obfuscating the `swagger.json` URL is as simple as writing a rule to forbid it (or hide if you prefer 404).

casparjespersen:
There can be an argument made that while hiding the documentation of API endpoints may not be considered as effective _security_, it still is a valid point in protecting business intelligence.

BenC14:
Is this going to be merged? Definitely agree there is a valid use case where you couldn't want to expose the swagger.json.

ravaiyapuri:
Hi

Is there an update on this? I'm looking for this enhancement as well.

PR #553, opened 2018-11-11
Add add_specs to api

alvarolopez:
This PR is basically #465 without the removal of the test.

Implements #464


casparjespersen:
What is status on this?

alvarolopez:
@casparjespersen the code is working, but it is pending to be reviewed/accepted by a maintainer.

j5awry:
I've approved, and will send it to the core team for a final check. Going through the ticket, it looks like @noirbizarre approved the general addition already. Code and tests look good to me. 

Sorry for not responding / checking earlier. Quite a backlog accumulated before more maintainers were added, and we're still going through things.

alvarolopez:
@j5awry no problems, thans for taking care of this.

SteadBytes:
This looks good to me, however before merging it would be worth adding to the documentation - perhaps under the [Hiding from documentation](https://flask-restplus.readthedocs.io/en/stable/swagger.html#hiding-from-documentation) section?

If you don't mind adding that, then I'm happy to merge this as @j5awry and @noirbizarre have already reviewed :+1:

alvarolopez:
@SteadBytes from my point of view the Swagger UI and the specs (i.e. swagger.json) are two different assets with different purposes that need to be handled separately. I agree on the documentation point though.

j5awry:
I agree with @alvarolopez on this. While the swagger GUI uses the generated swagger.json, they are distinct entities as far as end user cases. In that vein, I think we should ensure the other cases around `swagger.json` are discussed, tested, and we are clear on the outcomes:

1. Is there a case when hiding `swagger.json` from users is expected, but it is viewable in the GUI?
2. Does the current use of hiding `swagger.json` also remove it from the GUI?

SteadBytes:
@j5awry @alvarolopez , yes that's a good point. On second thoughts I agree with that. @j5awry, I can answer 2 for you; yes, the Swagger UI is not available if `swagger.json` is hidden. Visiting the URL for Swagger UI returns a `404`

alvarolopez:
@SteadBytes well, I guess this is a side-effect of removing the `swagger.json` that I find difficult to circumvent. I would just document that disabling specs removes the web UI as well.

SteadBytes:
Yes, removing `swagger.json` does mean that the Swagger UI shouldn't be shown. I don't think it makes sense to try and show an empty Swagger UI. As long as it is clearly documented that is the behaviour I would expect.

However I think @j5awry was also asking whether we should support a possible use case of showing the Swagger UI but not directly exposing `swagger.json`? IMO that isn't something we should try to support. It would also mean that we'd have to change how the Swagger UI retrieves the specs to build the page as they wouldn't be publicly available.

alvarolopez:
So, I have added the documentation and also disabled spec generation.

Issue #560, opened 2018-11-28
Can't install development dependencies with documentation instructions


hygorxaraujo:
According to the [contribution guide](https://flask-restplus.readthedocs.io/en/latest/contributing.html), to install the development dependencies we should run `pip install -e .[dev]`. When I execute the command I get the following warning: 

> flask-restplus 0.12.2.dev0 does not provide the extra 'dev'

Some packages are installed, but not everything listed in requirements/develop.pip. Here is the result of pip freeze:
```bash
$ pip freeze
aniso8601==4.0.1
Click==7.0
Flask==1.0.2
-e git+git@github.com:hygorxaraujo/flask-restplus.git@a8f35823fe40b2c7385632a2ad6b35b26467402c#egg=flask_restplus
itsdangerous==1.1.0
Jinja2==2.10
jsonschema==2.6.0
MarkupSafe==1.1.0
pytz==2018.7
six==1.11.0
Werkzeug==0.14.1
```
Steps to reproduce:
1. Fork flask-restplus repository in GitHub
2. Clone forked repository
3. Create new branch `git checkout -b new-branch`
4. Run `pip install -e .[dev]`

Am I missing something or is the documentation outdated?
Python: 3.6.7
Flask-Restplus: 0.12.2.dev

hygorxaraujo:
It seems that the problem is caused by a missing definition in the setuptools extras dict inside the `setup.py`.
Current:
```python
install_requires = pip('install')
doc_require = pip('doc')
tests_require = pip('test')

setup(...
    install_requires=install_requires,
    tests_require=tests_require,
    extras_require={
        'test': tests_require,
        'doc': doc_require,
    },
...)
```
Desired:
```python
install_requires = pip('install')
develop_requires = pip('develop')
doc_require = pip('doc')
tests_require = pip('test')

setup(...
    install_requires=install_requires,
    tests_require=tests_require,
    extras_require={
        'dev': develop_requires,
        'test': tests_require,
        'doc': doc_require,
    },
...)
```
But this seems to lead to another problem, running `pip install -e .[dev]` with the modified `setup.py` gives:
```
Complete output from command python setup.py egg_info:
    error in flask-restplus setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.
```
This is caused by the -e .[test] in `develop.pip`. A possible solution would be to change the pip function in `setup.py` to read the other requirements file and add the requirements to the list:
```python
[...]

PR #577, opened 2019-01-26
Issue 560 install dev dependencies

hygorxaraujo:
## Problem
As described in #560, it is not possible to install the development dependencies by following the instructions in the (documentation)[https://flask-restplus.readthedocs.io/en/stable/contributing.html]. Running the command `pip install -e .[dev]` leads to the following error:
>flask-restplus 0.12.2.dev0 does not provide the extra 'dev'

## Fix
To fix the problem, an extra `dev` was added to the `setup` function in `setup.py` and a small modification in the `pip` function (also in `setup.py`) to deal with the dependency between requirements files.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ Current
 -------
 
 - Ensure `basePath` is always a path
+- Fix `setup.py` to install `dev` dependencies (:issue:`560`)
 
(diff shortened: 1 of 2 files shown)


j5awry:
Closing PR. We accepted #618 for the implementation. This one kept the -e line, which is problematic.

PR #578, opened 2019-01-27
Use flask app logger for exception logging

Jaza:
This is a Flask extension, so it should log to the Flask app's logger, rather than defining its own logger.

In my Flask app, when `Swagger(self).as_dict()` throws an exception, I don't see the details of it in my app's log output, I guess I would need to specially configure the flask-restplus logger in order to see it, which would be unnecessarily tedious.

With this change, I see such exceptions in my app's log output, without any special setup needed.

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -4,7 +4,6 @@ from __future__ import unicode_literals
 import difflib
 import inspect
 from itertools import chain
-import logging
 import operator
 import re
 import six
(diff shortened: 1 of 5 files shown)


Jaza:
Note: my last commit "don't assume that params has an items attr", is needed in order for flask-marshmallow schemas (via flask-restplus-patched) to work with latest flask-restplus.

SteadBytes:
@Jaza Thank you for contributing, however Flask-RESTPlus should be defining it's own logger. This is best practice for writing extensions to Flask. Where is `Swagger(self).as_dict()` raising an error? Can you provide a code example? I'm more than happy to continue help debug the issue :smile:  However I am going to close this PR.

PR #580, opened 2019-01-28
-Fix ABC deprecations, targeted for removal in python 3.8

narfman0:
These changes avoid two warnings in python 3.7 similar to:
```
<venv>/lib/python3.7/site-packages/flask_restplus/swagger.py:8
  <venv>/lib/python3.7/site-packages/flask_restplus/swagger.py:8: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Hashable
```
Ostensibly, flask_restplus will break in python 3.8. It will fall back to old-style imports.

### DIFF
--- a/flask_restplus/reqparse.py
+++ b/flask_restplus/reqparse.py
@@ -4,7 +4,7 @@ from __future__ import unicode_literals
 import decimal
 import six
 
-from collections import Hashable
+from collections.abc import Hashable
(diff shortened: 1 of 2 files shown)


ziirish:
Thanks,

I will merge this now.

Issue #544, opened 2018-10-25
Add parameter config.SWAGGER_UI_OAUTH_REDIRECT_URL


aparamon:
Currently, Swagger has [oauth2RedirectUrl](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/#network) parameter always hard-coded to "http://localhost:3200/oauth2-redirect.html". It is sub-optimal: Flask-Restplus should respect `config.SWAGGER_UI_OAUTH_REDIRECT_URL` similar to `config.SWAGGER_UI_OAUTH_CLIENT_ID`, `config.SWAGGER_UI_OAUTH_APP_NAME`, `config.SWAGGER_UI_OAUTH_REALM`.

PR #585, opened 2019-02-04
Enable OAuth2 Implicit Flow authentication round-trip in Swagger UI

alanjcastonguay:
The swagger-ui project contains a 'oauth2-redirect.html' file
which provides a credential trampoline. Vendor it in, and
place the external url to this file in swagger-ui.html.

The OAuth2 authentication loop is run in a popup window/tab.
The IDP will redirect back to the oauth2RedirectUrl with an
access_token provided in the #hash-fragment of the url.
Javascript running inside 'oauth2-redirect.html' pushes the
access_token back to the parent which created the window/tab,
before closing the tab.

Enables use-case in #544 without adding another config param.

Usage:

```python
app.config.SWAGGER_UI_OAUTH_CLIENT_ID = 'MyClientId'
app.config.SWAGGER_UI_OAUTH_REALM = '-'
app.config.SWAGGER_UI_OAUTH_APP_NAME = 'Demo'
api = Api(
    app,
    title='Demo',
    security={'OAuth2': ['read', 'write']},
    authorizations={
        'OAuth2': {
            'type': 'oauth2',
            'flow': 'implicit',
            'authorizationUrl': 'https://example.auth0.com/authorize?audience=https://app.example.com',
            'clientId': Auth().client_id,
            'scopes': {
                'openid': 'Get ID token',
                'profile': 'Get identity',
            }
        }
    }
)
```

### DIFF
--- a/doc/swagger.rst
+++ b/doc/swagger.rst
@@ -889,6 +889,41 @@ You can specify a custom validator URL by setting ``config.SWAGGER_VALIDATOR_URL
     api = Api(app)
 
 
+You can enable [OAuth2 Implicit Flow](https://oauth.net/2/grant-types/implicit/) for retrieving an
+authorization token for testing api endpoints interactively within Swagger UI.
+The ``config.SWAGGER_UI_OAUTH_CLIENT_ID`` and ``authorizationUrl`` and ``scopes``
+will be specific to your OAuth2 IDP configuration.
+The realm string is added as a query parameter to authorizationUrl and tokenUrl.
+These values are all public knowledge. No *client secret* is specified here.
+.. Using PKCE instead of Implicit Flow depends on https://github.com/swagger-api/swagger-ui/issues/5348
+
+.. code-block:: python
+
+    from flask import Flask
+    app = Flask(__name__)
+
+    app.config.SWAGGER_UI_OAUTH_CLIENT_ID = 'MyClientId'
+    app.config.SWAGGER_UI_OAUTH_REALM = '-'
+    app.config.SWAGGER_UI_OAUTH_APP_NAME = 'Demo'
+
+    api = Api(
+        app,
+        title=app.config.SWAGGER_UI_OAUTH_APP_NAME,
+        security={'OAuth2': ['read', 'write']},
+        authorizations={
+            'OAuth2': {
+                'type': 'oauth2',
+                'flow': 'implicit',
+                'authorizationUrl': 'https://idp.example.com/authorize?audience=https://app.example.com',
+                'clientId': app.config.SWAGGER_UI_OAUTH_CLIENT_ID,
+                'scopes': {
+                    'openid': 'Get ID token',
+                    'profile': 'Get identity',
+                }
+            }
+        }
+    )
+
 You can also specify the initial expansion state with the ``config.SWAGGER_UI_DOC_EXPANSION``
 setting (``'none'``, ``'list'`` or ``'full'``):
 
--- a/flask_restplus/templates/swagger-ui.html
+++ b/flask_restplus/templates/swagger-ui.html
@@ -51,6 +51,9 @@
         window.onload = function() {
             const ui = window.ui = new SwaggerUIBundle({
                 url: "{{ specs_url }}",
+                {% if config.SWAGGER_UI_OAUTH_CLIENT_ID -%}
+                oauth2RedirectUrl: "{{ url_for('restplus_doc.static', filename='oauth2-redirect.html', _external=True) }}",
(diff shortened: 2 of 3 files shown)


alanjcastonguay:
@noirbizarre is this valuable? Does it look ok?

Colin-b:
Is it possible to get some feedback on this PR? Even if implicit flow was recently considered as unsafe (see latest RFC on this topic), this flow is still heavily used and not being able to use swagger-ui on implicit flow authenticated endpoints is a huge issue (swagger-ui is then useless for those endpoints).

Now that this project is back from the dead, I would really like for this PR (if it works) to make it in the next release.

Thanks again

alanjcastonguay:
@SteadBytes  can you take a look at this PR?

teuneboon:
Currently migrating an API to Flask + Restplus and running into this issue(Swagger-UI useless for any endpoints that use implicit auth). Any ETA on when this will be looked at? Want to know if I should start working on a workaround.

SteadBytes:
I do apologise for the delay in reviewing this PR and for responding @alanjcastonguay. It's been a busy few weeks and our lead maintainer @noirbizarre is currently taking a break due to the recent birth of his son :smile: so we are awaiting his return before making a new release to PyPI for the recent influx of new features/bug fixes that we have merged into master. I will take a look at this PR ASAP :+1:

SteadBytes:
Please correct me if I'm wrong, but I don't think that this addresses #544 :thinking: This feature will set the Swagger UI `oauth2RedirectUrl` configuration option to always reference `<static_path>/oauth2-redirect.html` if `config.SWAGGER_UI_OAUTH_CLIENT_ID` is set. #544 is asking for a way to configure `oauth2RedirectUrl` to point to *any* URL for OAuth redirects. This allows usage of the OAuth implicit authentication flow but only using `<static_path>/oauth2-redirect.html` which is a different feature I think?

alanjcastonguay:
This was several month ago so my memory is a bit lacking. But iirc the arbitrary Other url (for an OAuth2 IDP) is configured with `authorizationUrl` like in the example above. The HTML file here is a trampoline.

alanjcastonguay:
> @noirbizarre is currently taking a break due to the recent birth of his son

Woot!

SteadBytes:
@alanjcastonguay Ah yes that is correct thank you for clarifying. The code changes look good to me :+1: It would however be great if some documentation could be added for this - do you mind doing that? I will also ask for review from some fellow contributors :smile:

alanjcastonguay:
@SteadBytes I pushed 7d5795b with some docs. Is that what you're looking for?

doc/swagger.rst
```diff
@@ -889,6 +889,41 @@ You can specify a custom validator URL by setting ``config.SWAGGER_VALIDATOR_URL
     api = Api(app)
 
 
+You can enable [OAuth2 Implicit Flow](https://oauth.net/2/grant-types/implicit/) for retrieving an
+authorization token for testing api endpoints interactively within Swagger UI.
+The ``config.SWAGGER_UI_OAUTH_CLIENT_ID`` and ``authorizationUrl`` and ``scopes``
+will be specific to your OAuth2 IDP configuration.
+The realm string is added as a query parameter to authorizationUrl and tokenUrl.
+These values are all public knowledge. No *client secret* is specified here.
+.. Using PKCE instead of Implicit Flow depends on https://github.com/swagger-api/swagger-ui/issues/5348
```
SteadBytes: I don't think it's a good idea to link an unresolved issue in the user documentation. I understand the reasoning behind it, however as it is not currently a feature, including this link makes the documentation more confusing. Furthermore, I can see it easily being forgotten - documentation is always hard to keep up to date (I've forgotten items like this many times :joy: ). Would you mind removing this line, then we can get this PR merged in. Thank you again!

SteadBytes:
@alanjcastonguay Thank you for including the documentation - it is much appreciated! I have left a single comment in a review on it. If you're happy to address that I'll get this merged :+1:

doc/swagger.rst
alanjcastonguay: I intended this to be a comment (dot dot space arbitrary text) in the documentation source, explicitly to help future maintainers deal with such. Unless I am mistaken as to how comments work, this is not rendered in the consumer-visible result.

Does that change your feedback?

SteadBytes: Oh yes sorry, I apologise! I was not reading carefully enough :thinking:  I'm still concerned that the comment will get forgotten so I will link to this PR on that issue such that it's visible for when that issue is resolved. Thank you again !

PR #588, opened 2019-02-06
Suggestion: Add support for streaming

kurt-hectic:
I have implemented support for http streaming using the flask Response object. This can be useful when generating large responses as it avoids timeouts.
In my case it has helped me to get around an annoying 30 seconds timeout on Heroku, while still being able to use the restplus object marshaling.
Additional decorators could be added if people find it useful.


dnaeon:
What's the status of this PR? Any chance this gets merged soon?

PR #596, opened 2019-02-26
Feature/swagger ui behind proxy

adroffner:
This branch adds a **behind_proxy=True** keyword parameter to the **flask_restplus.api.Api(..., behind_proxy=True)** object. Many users need a _reverse proxy_ like **nginx** to serve the REST API. In my case, I had to place the server inside a kubernetes **pod** & keep its **port** number.

I added appropriate unit tests to the **test_api.py** file. It considers both cases when **behind_proxy** is either **True** or **False**.  I assume that **SERVER_NAME** is set & has a defined **port**.

OTHER ISSUES: I did not want to fix these in this branch & make it change too much

* I had to `pip install wheel` to make the `invoke dist` work, which is not listed in the **requirements** directory. 
* I had to run `invoke assets` _before_ `invoke dist` so that the **Swagger UI** javascript gets into the **wheel**
* `invoke qa` shows many **mccabe complexity** violations (C901 flake8 warnings) that stop the build

The **C901 complexity** warnings mean no one can just run `inv all` to complete a build :(

### DIFF
--- a/flask_restplus/__about__.py
+++ b/flask_restplus/__about__.py
@@ -1,3 +1,3 @@
 # -*- coding: utf-8 -*-
-__version__ = '0.12.2.dev-behind_proxy'
+__version__ = '0.12.3.dev'
 __description__ = 'Fully featured framework for fast, easy and documented API development with Flask'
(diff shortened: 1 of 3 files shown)


michaelbukachi:
I really hope this pull request is merged. There's hack I keep copying from project to project just to apps work behind proxies.😂

PR #597, opened 2019-02-28
fix list field accept dict parameter bug

Friday21:
when I use List field with nested object field, I find out when the client pass the object filed(not as list), the parameter check didn't throw parameter error(400), so I got a json object in my view layer, that is not what I expected,  thus trigger a bug。

### DIFF
--- a/flask_restplus/fields.py
+++ b/flask_restplus/fields.py
@@ -273,8 +273,11 @@ class List(Raw):
 
     def format(self, value):
         # Convert all instances in typed list to container type
-        if isinstance(value, set):
+        if isinstance(value, (set, tuple)):
             value = list(value)
+
+        if not isinstance(value, list):
+            raise MarshallingError('should be list')
 
         is_nested = isinstance(self.container, Nested) or type(self.container) is Raw
 
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -836,6 +836,13 @@ class ListFieldTest(BaseFieldTestMixin, FieldTestCase):
         expected = [OrderedDict([('name', 'John Doe')]), OrderedDict([('name', 'Jane Doe')])]
         self.assert_field(field, data, expected)
 
+    def test_format_error(self, api):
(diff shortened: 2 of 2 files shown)


SteadBytes:
Thanks for contributing! Could you please provide some example code to reproduce the issue in the same way you encountered it?

Friday21:
> Thanks for contributing! Could you please provide some example code to reproduce the issue in the same way you encountered it?

I think the test case I add in test_format_error has already include one example

SteadBytes:
I see the test case, however it's not a full example of the use case that you described. Calling the `format` method on a field directly is very different to how it is used when receiving a POST payload

Friday21:
> I see the test case (thank you for that), however it's not a full example of the use case that you described. Calling the `format` method on a field directly is very different to how it is used when receiving a POST payload

OK, so here is my example in real project(change some name to hide work info):

first, the input model definition:
```
time_fields = api.model('TimeInput', {
    'id': fields.Integer(required=True, description='ID'),
    'time_area_list': fields.List(fields.List(fields.Integer), required=True,
                                  description='time area, ex： [[t1, t2], [t3, t4], [t5], [t6, t7]...]')
})

test_list_fields = api.model('TimelistInput', {
    'list_time': fields.List(fields.Nested(time_fields), required=True),
})
```

and, here is the code in the rest api code:
```
@time_api.route('/beat', methods=['POST', 'PUT'])
class TimeResource(Resource):
    @time_api.expect(input.test_list_fields)
    def post(self):
        """"""
        payload = self.api.payload
        time_list = payload.get('list_time')
        print(time_list)
        return response(ResponseCode.SUCCESS, msg='success')

```

and then, I send request through postman with list_time field as list of object

Screen Shot 2019-03-31 at 10 21 01 PM


this works as expected. then I send list_time field as object: 

Screen Shot 2019-03-31 at 10 21 46 PM

it should have return parameter error, but it didn't, it just pass on. the print info is below:

Screen Shot 2019-03-31 at 10 22 16 PM

Issue #565, opened 2018-12-18
applying nginx with ssl resulting swagger says No API definition provided.


bilalbayasut:
I have implemented nginx and ssl , but when I run the swagger UI it says No API definition provided

bilalbayasut:
Fixing this by do monkey patching 
```
# monkey patch courtesy of
# https://github.com/noirbizarre/flask-restplus/issues/54
# so that /swagger.json is served over https
if os.environ.get('HTTPS'):
    @property
    def specs_url(self):
        """Monkey patch for HTTPS"""
        return url_for(self.endpoint('specs'), _external=True, _scheme='https')

    Api.specs_url = specs_url

```
hope it helps

livecloud:
Curious, which release will include this fix?

ibt23sec5:
Use ProxyFix:
```
from werkzeug.middleware.proxy_fix import ProxyFix

app = Flask(__name__)
app.wsgi_app = ProxyFix(app.wsgi_app)
```

ibt23sec5:
Use proxyFix:

```
from werkzeug.contrib.fixers import ProxyFix
app = Flask(__name__)
app.wsgi_app = ProxyFix(app.wsgi_app)
```

SteadBytes:
Yep @ibt23sec5 has the answer 👍

MartinThoma:
Is this still an issue in https://github.com/python-restx/flask-restx or was it fixed there?

ziirish:
As discussed several times, I don't think this is an issue with Flask-restplus/restx as the werkzeug's ProxyFix is here to handle such case.

So no, for now it's not "fixed" in flask-restx either.

Now if someone has an actual problem/bug with this, I'm okay to discuss a change in flask-restx.
Feel free to open an issue with as much details as you can.

PR #599, opened 2019-03-01
auto judge scheme of specs_url

adolli:
fix API specs page MixedContentError when served over https.

image

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -451,7 +451,8 @@ class Api(object):
 
         :rtype: str
         '''
-        return url_for(self.endpoint('specs'), _external=True)
+        _scheme = request.headers.get('x-forwarded-proto', 'http')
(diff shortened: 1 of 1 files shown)


flask_restplus/api.py
```diff
@@ -451,7 +451,8 @@ def specs_url(self):
 
         :rtype: str
         '''
-        return url_for(self.endpoint('specs'), _external=True)
+        _scheme = request.headers.get('x-forwarded-proto', 'http')
+        return url_for(self.endpoint('specs'), _external=True, _scheme=_scheme)
```
yanirs: I just hit the same issue and resolved it with monkey patching `_scheme=''`, similarly to the suggestion in #565. What's the advantage of using the approach in this PR versus setting `_scheme=''` to get a protocol-relative URL?

PR #600, opened 2019-03-04
Fix a problem of operation expand.

hayatesa:
Several operation tabs would be expanded by one expand operation in doc page when a namespace includes duplicate http methods, since they have the same operationId. So, I join it a random string generated by "str(uuid.uuid1())[:6]".


SteadBytes:
Thanks for contributing! This PR is causing a lot of failed unit tests - are you willing to look into and resolve this? Could you also provide an example to reproduce the issue please?

PR #602, opened 2019-03-06
solve error if method view's docstring is chinese

zhongpengqun:
My docstring is chinese, so i add those 3 lines code to avoid it reporting error


SteadBytes:
@zhongpengqun Thanks for contributing, this PR is failing quite a few unit tests, are you willing to look into and fix this?

Issue #609, opened 2019-03-22
@api.expect() decorator with "validate=False" doesn't work


dstrigl:
Corresponding to the documentation at [https://flask-restplus.readthedocs.io/en/stable/swagger.html](https://flask-restplus.readthedocs.io/en/stable/swagger.html) the following sample
```
api = Api(app, validate=True)
resource_fields = api.model('Resource', {
    'name': fields.String,
})
@api.route('/my-resource/<id>')
class MyResource(Resource):
    # Payload validation disabled
    @api.expect(resource_fields, validate=False)
    def post(self):
        pass`
```
doesn't work as expected:

The payload validation for the POST method is active, although the decorator `@api.expect(...)` is called with `validate=False`.

After a short look in the source code I would say that both, the function `expect` in namespace.py and the unit test `test_validation_true_on_constructor_with_override` in test_payload.py is not really correct.

xshibux:
I'm having the same issue.  Can someone help?

dstrigl:
[https://github.com/noirbizarre/flask-restplus/pull/610](url)

SteadBytes:
You are correct, the intended behaviour is for validation on all resources to default to the validation set on the `Api` constructor but can be overridden on a per resources basis by passing the `validate` kwarg to `api.expect` as show in your demo @dstrigl . I am going to take a look at your PR now :+1:

SteadBytes:
@dstrigl I have left a comment on the PR :smile:

ibejohn818:
Since this hasn't been merged into an upstream release there is another work around with the current release. You can use the "doc" decorator to get the same "payload" documentation and disable validation
IE:
`@doc(body=my_model, validate=False)`

PR #610, opened 2019-03-22
Fix @api.expect() decorators with validate=False (issue-609)

dstrigl:
See issue #609

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Changelog
 Current
 -------
 
+- Fix `@api.expect(..., validate=False)` decorators for an :class:`Api` where `validate=True` is set on the constructor
 - Ensure `basePath` is always a path
 
 0.12.1 (2018-09-28)
--- a/flask_restplus/namespace.py
+++ b/flask_restplus/namespace.py
@@ -203,7 +203,7 @@ class Namespace(object):
         '''
         expect = []
         params = {
-            'validate': kwargs.get('validate', None) or self._validate,
+            'validate': kwargs.get('validate', self._validate),
             'expect': expect
         }
         for param in inputs:
--- a/tests/test_payload.py
+++ b/tests/test_payload.py
@@ -79,7 +79,8 @@ class PayloadTest(object):
             def post(self):
                 return {}
 
-        self.assert_errors(client, '/validation/', {}, 'name')
+        data = client.post_json('/validation/', {})
+        assert data == {}
 
     def _setup_api_format_checker_tests(self, app, format_checker=None):
         class IPAddress(restplus.fields.Raw):
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -592,6 +592,13 @@ class Api(object):
         '''
         got_request_exception.send(current_app._get_current_object(), exception=e)
 
+        if not isinstance(e, HTTPException) and current_app.propagate_exceptions:
+            exc_type, exc_value, tb = sys.exc_info()
+            if exc_value is e:
+                raise
+            else:
+                raise e
+
         include_message_in_response = current_app.config.get("ERROR_INCLUDE_MESSAGE", True)
         default_data = {}
 
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -291,12 +291,13 @@ class ErrorsTest(object):
 
         app.config['PROPAGATE_EXCEPTIONS'] = True
 
-        response = client.get('/api/test/')
-        assert response.status_code == 500
-        assert response.content_type == 'application/json'
+        # From the Flask docs:
+        # PROPAGATE_EXCEPTIONS
+        # Exceptions are re-raised rather than being handled by the app’s error handlers.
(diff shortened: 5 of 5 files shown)


flask_restplus/namespace.py
```diff
@@ -201,9 +201,10 @@ def expect(self, *inputs, **kwargs):
         :param bool validate: whether to perform validation or not
 
         '''
+        validate = kwargs.get('validate', None)
         expect = []
         params = {
-            'validate': kwargs.get('validate', None) or self._validate,
+            'validate': validate if validate is not None else self._validate,
```
SteadBytes: There is no need to separate out retrieving `validate` from `kwargs` as the default parameter for `dict.get` can use `self._validate`:
```python
kwargs.get('validate', self._validate)
```

tests/test_payload.py
```diff
@@ -79,7 +79,8 @@ class ValidationOff(restplus.Resource):
             def post(self):
                 return {}
 
-        self.assert_errors(client, '/validation/', {}, 'name')
```
SteadBytes: Great spot on this unit test being incorrect :+1:

CHANGELOG.rst
```diff
@@ -6,6 +6,7 @@ Changelog
 Current
 -------
 
+- Fix `@api.expect()` decorators with `validate=False` (:issue:`609`)
```
SteadBytes: It would be clearer to include that this is for when `validate=True` is set on the `Api` constructor :smile:

flask_restplus/namespace.py
dstrigl: Oh, you are right ... much easier to read!

tests/test_payload.py
dstrigl: :grin:

CHANGELOG.rst
dstrigl: Hard to formulate in one short sentence.
```suggestion
- Fix `@api.expect(..., validate=False)` decorators for an :class:`Api` where `validate=True` is set on the constructor
```
Better?

SteadBytes:
Looks good @dstrigl , could you just rebase onto master and squash your last 'some enhancement' commit (2d3a666) please? Then I will merge :+1:

dstrigl:
Hi @SteadBytes, I hope I did it the right way ... I am not yet so versed with the github workflow.

SteadBytes:
@dstrigl not quite sorry I may not have been clear enough. What we want to achieve is cleaning up the commit history so the 'some enhancement' commit (2d3a666) is not shown. To do this we can do an interactive rebase and squash the commit into the previous one:

Make sure your master branch is up to date:
`git checkout master && git pull`

Checkout your feature branch:
`git checkout dstrigl:issue-609`

Start an interactive rebase:
`git rebase -i master`

This should drop you into a text editor listing each commit with the work `pick` at the beginning of each line. To squash the commit into the previous one and remove it's message, change the word `pick` at the start of the line for the 'some enhancement' commit to `fixup`. Save and exit, the rebase will be applied. Then force push your branch.

I hope that helps 😊

dstrigl:
Hi @SteadBytes, thanks for explaining the workflow in more detail ...
So the reason behind is that you would like to have a more cleaner commit history (and remove some unnecessary commits, like the "some enhancement")?

... hope I did it right now!

SteadBytes:
Yes @dstrigl that's the idea - we want the changes from the commits but not the commit messages cluttering up the history 😀 

Unfortunately you seem to have introduced some commits from somewhere else - if you have a look at the commits for this PR there are now some from other people 🤔

Did you make sure your master branch was up to date first?

dstrigl:
Hi @SteadBytes, I admit I was a bit confused too why there are commits from other peoples. But I performed exactly the steps you proposed me. Sorry for the mistakes ... I never used rebase before, we normally work with merge-commits ...

SteadBytes:
It's always worth checking the commit history before pushing :+1: So this will still use a merge commit when this branch is merged into master. The rebase is used to clean up the commit  history before that. 

I've just cloned your repo and the problem is that your forked `master` branch was not up to date with the upstream `master` in the main Flask RESTPlus repo. Here's the fix:

Checkout your master branch:
`git checkout master`

Add an upstream for Flask RESTPlus:
```
git remote add upstream https://github.com/noirbizarre/flask-restplus.git
git fetch upstream
```
Update your `master` branch from the new upstream:
`git pull upstream master`

Checkout your feature branch and merge in master:
```
git checkout issue-609
git merge master
```

If all went as expected, the following should only show your commit:
`git diff upstream/master`

If that worked, push your branch again and that should clear up the commits

dstrigl:
Hi @SteadBytes,

> I've just cloned your repo and the problem is that your forked master branch was not up to date with
> the upstream master in the main Flask RESTPlus repo.

that was the point ... I forgot to add an upstream ...  (_shame on me_)

Now I hope that was the last correction :worried: and thanks in tutoring me in this git workflow ...

SteadBytes:
Awesome - I will merge the PR now 😁

You are more than welcome, I'm glad to help 👍

PR #616, opened 2019-03-30
 fix missing definition of nested ordered model

illes:
Example UI error message:
```
Could not resolve pointer: /definitions/XXX does not exist in document
```

### DIFF
--- a/flask_restplus/swagger.py
+++ b/flask_restplus/swagger.py
@@ -12,7 +12,7 @@ from flask import current_app
 from werkzeug.routing import parse_rule
 
 from . import fields
-from .model import Model, ModelBase
+from .model import Model, ModelBase, OrderedModel
 from .reqparse import RequestParser
 from .utils import merge, not_none, not_none_sorted
 from ._http import HTTPStatus
@@ -46,7 +46,8 @@ RE_RAISES = re.compile(r'^:raises\s+(?P<name>[\w\d_]+)\s*:\s*(?P<description>.*)
 
 def ref(model):
     '''Return a reference to model in definitions'''
-    name = model.name if isinstance(model, ModelBase) else model
+    name = model.name if 
+    nce(model, ModelBase) else model
     return {'$ref': '#/definitions/{0}'.format(name)}
 
 
@@ -537,7 +538,7 @@ class Swagger(object):
         if isinstance(specs, ModelBase):
             for parent in specs.__parents__:
                 self.register_model(parent)
-        if isinstance(specs, Model):
+        if isinstance(specs, (Model, OrderedModel)):
             for field in itervalues(specs):
                 self.register_field(field)
         return ref(model)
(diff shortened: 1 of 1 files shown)


SteadBytes:
Thank you for contributing! Is there an issue that goes along with this PR? If not, could please include some steps to reproduce the issue and to demonstrate the fix?

illes:
Minimal example:
```python
import json

from flask import Flask
from flask_restplus import fields, Api, Swagger, Resource

app = Flask(__name__)
app.config['SERVER_NAME'] = "localhost"

api = Api(app, ordered=True)

simple = api.model('Simple', [
    ('foo', fields.String),
])

composite = api.model('Composite', [
    ('simple', fields.Nested(model=simple)),
    ('bar', fields.String),
])

@api.route('/')
class Composite(Resource):
    @api.marshal_with(composite)
    def get(self):
        pass

with app.app_context():
    swagger_json = Swagger(api).as_dict()
    print(json.dumps(swagger_json, indent=2, sort_keys=True))
```
generates the following `swagger.json` :
```json
{
  "basePath": "http://localhost", 
  "consumes": [
    "application/json"
  ], 
  "definitions": {
    "Composite": {
      "properties": {
        "bar": {
          "type": "string"
        }, 
        "simple": {
          "$ref": "#/definitions/Simple"
        }
      }, 
      "type": "object"
    }
  }, 
  "host": "localhost", 
  "info": {
    "title": "API", 
    "version": "1.0"
  }, 
  "paths": {
    "/": {
      "get": {
        "operationId": "get_composite", 
        "parameters": [
          {
            "description": "An optional fields mask", 
            "format": "mask", 
            "in": "header", 
            "name": "X-Fields", 
            "type": "string"
          }
        ], 
        "responses": {
          "200": {
            "description": "Success", 
            "schema": {
              "$ref": "#/definitions/Composite"
            }
          }
        }, 
        "tags": [
          "default"
        ]
      }
    }
  }, 
  "produces": [
    "application/json"
  ], 
  "responses": {
    "MaskError": {
      "description": "When any error occurs on mask"
    }, 
    "ParseError": {
      "description": "When a mask can't be parsed"
    }
  }, 
  "swagger": "2.0", 
  "tags": [
    {
      "description": "Default namespace", 
      "name": "default"
    }
  ]
}
```
which is lacking the definition of the type `Simple` which is an `OrderedModel` and only appears in a nested field:
```json
...
    "Simple": {
      "properties": {
        "foo": {
          "type": "string"
        }
      },
      "type": "object"
    }
...
```
the commits in the PR fix this

arabidopsis:
Can confirm that I'm having this problem too and have had to monkeypatch flask_restplus
with basically this pull request.

j5awry:
Going to check this and also check against cases listed in pr #640 and the issues linked there. It seems to have similar characteristics, though with the wide variety of symptoms we're seeing against nested schema, it's making it hard to focus it down.

j5awry:
Was able to reproduce specific issue regarding Ordered. Also confirm this patch fixes the specific problem of ordered APIs and using api.model schema style generation.

This doesn't help with JSON schema related issues. 

Being said, definitely worth bringing in for fixing ordered APIs.

j5awry:
This also addresses some cases found in issues #643 . We should pull this in and ensure it goes out in our upcoming release

alvarolopez:
Hi.

Any idea when this will get merged?

Issue #587, opened 2019-02-05
ujson segfaults with numpy arrays


jackieleng:
In `flask_restplus.representations:output_json` (https://github.com/noirbizarre/flask-restplus/blob/master/flask_restplus/representations.py#L25)
`usjon.dumps` is called (when installed) with `indent=4` when DEBUG is enabled. However this is causing a segfault when you're trying to serialize a numpy array. Without indent it also causes an error, but it's an `OverflowError`. See for example this issue: https://github.com/esnme/ultrajson/issues/294

I know it's a problem with `ujson`,  but it does kill the server and it's not easy to find the cause.

ziirish:
I think we will probably drop `ujson` as stated in #589 because the project does not seem actively supported anymore.

Thanks for the report.

Issue #507, opened 2018-08-03
A TypeError happened with ujson.dumps imported and DEBUG set True


furoxr:
With app.debug set True, and the module ujson imported, When I accessed the doc of api, a TypeError was raised.

In the **representations.py**, the code of importing dumps
```
from __future__ import unicode_literals, absolute_import

try:
    from ujson import dumps
except ImportError:
    from json import dumps

from flask import make_response, current_app
```

and the function output_json in the same file
```
def output_json(data, code, headers=None):
    '''Makes a Flask response with a JSON encoded body'''

    settings = current_app.config.get('RESTPLUS_JSON', {})

    # If we're in debug mode, and the indent is not set, we set it to a
    # reasonable value here.  Note that this won't override any existing value
    # that was set.
    if current_app.debug:
        settings.setdefault('indent', 4)

    # always end the json dumps with a new line
    # see https://github.com/mitsuhiko/flask/pull/1262
    dumped = dumps(data, **settings) + "\n"

    resp = make_response(dumped, code)
    resp.headers.extend(headers or {})
    return resp

```
ujson.dumps() accepts one arg, while json.dumps() accepts one or more. So, when the app.debug is true,  dumps will accept two args, 'data' and 'settings', with` settings = { 'indent' : 4 }`. 

Then `TypeError: 'indent' is an invalid keyword argument for this function`

ziirish:
Hopefully this will be addressed with #589

PR #637, opened 2019-05-06
stop relying on ujson (fix #589, #587, #507)

### LINKED ISSUE #589: Stop using failing ujson as default serializer
ujson is no more actively maintained and specially for some critical bugs not present in json package.

one example forcing us to revert ujson usage:
https://github.com/esnme/ultrajson/issues/325 (2 days ago, mine)
https://github.com/esnme/ultrajson/issues/301 (same but 1 year ago and no answer)

ziirish:
ujson does not seem to be maintained anymore. Some bugs reported months (or years ago) start to affect our users.
Since we don't ship with ujson in our requirements and this was just a "optional performance improvement" I think it is safe to remove it.

Note: this reverts #318

### DIFF
--- a/flask_restplus/representations.py
+++ b/flask_restplus/representations.py
@@ -1,10 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals, absolute_import
 
-try:
-    from ujson import dumps
-except ImportError:
-    from json import dumps
+from json import dumps
 
 from flask import make_response, current_app
 


ziirish:
Coverall seems drunk.
It noticed there is one line less than in the previous commit dropping the number of covered lines from 11 to 10... right, but it did not notice there are several lines less than in the previous commit.

ziirish:
As discussed with @noirbizarre we give people the ability to choose their own json serializer through a new dedicated `RESTPLUS_JSON_SERIALIZER` config option.
The default being to use python's build-in serializer by default.

This way if they are not affected by a `ujson` bug they can still use it as an opt-in option, or they can even use whatever serializer they like (there are plenty nowadays)

ziirish:
@j5awry actually the import only occurs once (during the first call) then the serializer is *cached* within the `serializer` global variable.

I just ran some tests to confirm the caching is working as it should and it turns out it does =)

j5awry:
> @j5awry actually the import only occurs once (during the first call) then the serializer is _cached_ within the `serializer` global variable.
> 
> I just ran some tests to confirm the caching is working as it should and it turns out it does =)

Sweet! i'd still like it as a separate module for clarity, but that's a personal thing. If it sticks around cached in the global, then we're good! Again a personal thing with an aversion to global :)

SteadBytes:
Does using the default Python `json` module have any performance implications over the current usage of `ujson`?

ziirish:
I don't know whether this is representative or not but here are the benchmark results.

Without ujson:
```
>> Run benchmarks
Test session starts (platform: linux, Python 3.6.8, pytest 3.2.3, pytest-sugar 0.9.0)
benchmark: 3.1.1 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=2 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /opt/workspace/flask-restplus, inifile: setup.cfg
plugins: sugar-0.9.0, profiling-1.2.11, mock-1.6.3, flask-0.10.0, faker-2.0.0, cov-2.5.1, benchmark-3.1.1

 tests/benchmarks/bench_marshalling.py ✓✓✓✓                                                                                                                                                                                      67% ██████▋   
 tests/benchmarks/bench_swagger.py ✓✓                                                                                                                                                                                           100% ██████████

------------------------------------------------------------------------------------------ benchmark 'marshalling': 4 tests -----------------------------------------------------------------------------------------
Name (time in us)                         Min                    Max                  Mean              StdDev                Median                IQR            Outliers         OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_marshal_simple                 429.9954 (1.0)       2,721.3283 (1.44)       447.7427 (1.0)       49.8622 (1.0)        447.1317 (1.0)      13.0665 (1.0)        16;106  2,233.4255 (1.0)        2220           1
bench_marshal_simple_with_mask     1,173.6453 (2.73)      1,895.1111 (1.0)      1,216.0383 (2.72)      72.7597 (1.46)     1,200.3686 (2.68)     21.7594 (1.67)        54;70    822.3425 (0.37)       1396           1
bench_marshal_nested               1,844.3316 (4.29)      2,925.9734 (1.54)     1,885.2151 (4.21)      49.8984 (1.00)     1,877.7251 (4.20)     26.5837 (2.03)        62;74    530.4434 (0.24)       1021           1
bench_marshal_nested_with_mask     2,606.2354 (6.06)     20,920.8205 (11.04)    2,762.5558 (6.17)     783.2512 (15.71)    2,675.0267 (5.98)     71.7286 (5.49)         8;79    361.9836 (0.16)        736           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------------------ benchmark 'swagger': 2 tests -----------------------------------------------------------------------------------------
Name (time in us)                     Min                   Max                  Mean              StdDev                Median                 IQR            Outliers         OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_swagger_specs_cached       732.1090 (1.0)      1,028.4223 (1.0)        753.2495 (1.0)       36.4551 (1.0)        738.8182 (1.0)       18.8295 (1.0)         30;33  1,327.5814 (1.0)         276           1
bench_swagger_specs            6,795.6112 (9.28)     7,992.8115 (7.77)     6,943.3125 (9.22)     143.5110 (3.94)     6,859.9358 (9.29)     217.2068 (11.54)        36;4    144.0235 (0.11)        342           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean

Results (12.57s):
       6 passed
```

With `app.config['RESTPLUS_JSON_SERIALIZER'] = 'ujson'`:
```
>> Run benchmarks
Test session starts (platform: linux, Python 3.6.8, pytest 3.2.3, pytest-sugar 0.9.0)
benchmark: 3.1.1 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=2 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /opt/workspace/flask-restplus, inifile: setup.cfg
plugins: sugar-0.9.0, profiling-1.2.11, mock-1.6.3, flask-0.10.0, faker-2.0.0, cov-2.5.1, benchmark-3.1.1

 tests/benchmarks/bench_marshalling.py ✓✓✓✓                                                                                                                                                                                      67% ██████▋   
 tests/benchmarks/bench_swagger.py ✓✓                                                                                                                                                                                           100% ██████████

----------------------------------------------------------------------------------------- benchmark 'marshalling': 4 tests -----------------------------------------------------------------------------------------
Name (time in us)                         Min                   Max                  Mean              StdDev                Median                IQR            Outliers         OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_marshal_simple                 431.2173 (1.0)      1,028.7687 (1.0)        443.5333 (1.0)       16.7665 (1.0)        441.4711 (1.0)       8.2292 (1.0)       187;235  2,254.6223 (1.0)        2248           1
bench_marshal_simple_with_mask     1,174.5617 (2.72)     2,034.2432 (1.98)     1,216.6918 (2.74)      79.1666 (4.72)     1,196.5148 (2.71)     23.9285 (2.91)       74;111    821.9008 (0.36)       1377           1
bench_marshal_nested               1,799.6989 (4.17)     2,629.2317 (2.56)     1,847.2305 (4.16)      50.8462 (3.03)     1,838.7847 (4.17)     26.0845 (3.17)        56;63    541.3510 (0.24)       1026           1
bench_marshal_nested_with_mask     2,561.9529 (5.94)     4,902.0723 (4.76)     2,656.1316 (5.99)     144.1454 (8.60)     2,619.2032 (5.93)     51.0439 (6.20)        45;66    376.4874 (0.17)        643           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------------------- benchmark 'swagger': 2 tests -------------------------------------------------------------------------------------------
Name (time in us)                     Min                    Max                  Mean                StdDev                Median                 IQR            Outliers         OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_swagger_specs_cached       730.5555 (1.0)         789.5790 (1.0)        738.7847 (1.0)          8.5986 (1.0)        735.3090 (1.0)        3.2894 (1.0)         52;53  1,353.5743 (1.0)         284           1
bench_swagger_specs            6,796.8778 (9.30)     43,136.9580 (54.63)    7,136.4213 (9.66)     1,976.2746 (229.84)   6,988.0933 (9.50)     234.0488 (71.15)        1;28    140.1263 (0.10)        339           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean

Results (11.57s):
       6 passed
```

SteadBytes:
Looks like for the average case (mean) the two are fairly similar, however `ujson` has a *much* smaller standard deviation for the marshalling benchmarks - interesting! I'm glad it's not a horrendous slow down or anything :sweat_smile:, so I'm happy with dropping `ujson` default.

On a separate note, I have just noticed the addition of the configuration options documentation - nice :+1: Would you mind including the `SWAGGER_SUPPORTED_SUBMIT_METHODS` option that was merged the other day too?

flask_restplus/representations.py
```diff
@@ -1,18 +1,32 @@
...
 
+    global serializer
+
     settings = current_app.config.get('RESTPLUS_JSON', {})
+    custom_serializer = current_app.config.get('RESTPLUS_JSON_SERIALIZER', None)
+
+    # If the user wants to use a custom serializer, let it be
+    if serializer is None and custom_serializer:
+        try:
+            serializer = importlib.import_module("{}.dumps".format(custom_serializer) if '.' not in custom_serializer
```
SteadBytes: This import is failing for me, for example setting `RESTPLUS_JSON_SERIALIZER='ujson'`  causes a `ModuleNotFoundError`. This can be replicated as such:

```python
import importlib

importlib.import_module("ujson.dumps")
# ModuleNotFoundError: No module named 'ujson.dumps'; 'ujson' is not a package
```

This because `ujson.dumps` is a function, not a module. A solution could be:
```python
import importlib

if serializer is None and custom_serializer is not None:
        try:
            # user provided a module path, attempt to import the 'dumps' method of the module
            mod = importlib.import_module(custom_serializer)
            if hasattr(mod, "dumps"):
                serializer = getattr(mod, "dumps")
        except ImportError:   
             if "." in custom_serializer:
                # user may have provided a path to a function
                # attempt to import the module and retrieve the function attribute
                parts = custom_serializer.split(".")
                mod_path = "".join(parts[:-1])
                func_name = parts[-1]
                try:
                    mod = importlib.import_module(mod_path)
                    if hasattr(mod, func_name):
                        serializer = getattr(mod, func_name)
                except ImportError:
                    pass
    elif serializer is None:
        serializer = dumps
```

Is it also worth logging a warning if the import fails so the user knows their custom serializer isn't being imported?

ziirish: Thanks for your careful review @SteadBytes!
I believe I have now fixed the issue and I added a few more tests to make sure the imported *serializer* is the expected one.

I also added a note in the doc stating that we will always silently fallback to `json.dumps` if we cannot manage to import the configured one.

We might add a more explicit warning thrown by the `output_json` func in the future.

tests/test_representations.py
```diff
@@ -0,0 +1,33 @@
...
+        'address': 'test',
+    }
+    r = rep.output_json(payload, 200)
+    assert loads(r.get_data(True)) == loads(dumps(payload))
+    # now reset serializer
+    rep.serializer = None
+    # then enforce a custom serializer
+    current_app.config['RESTPLUS_JSON_SERIALIZER'] = 'ujson'
+    r2 = rep.output_json(payload, 200)
+    assert uloads(r2.get_data(True)) == uloads(udumps(payload))
```
SteadBytes: This is testing several different things within a single test case - how do you feel about splitting it up into several test cases so that it's clearer to see the individual assertions? Something like:
```python
def test_representations_serialization_output_correct():
    # original test comparing get_data and dumps outputs

def test_config_custom_serializer_is_module():
    # current_app.config['RESTPLUS_JSON_SERIALIZER'] = 'ujson'
    # test .dumps method is used 

def test_config_custom_serializer_is_function():
    # current_app.config['RESTPLUS_JSON_SERIALIZER'] = 'my_custom_json.module.dump_method'
    # test correct function imported

def test_config_custom_serializer_fallback():
    # test fallback to json.dumps
```
This makes the intended behaviour clearer and is easier to see from a test failure where the issue is :smile:

ziirish: Thanks!

I have spit the test.
I'm not quite sure about the `warnings.warn` yet though. I think we should define a *communication with upper apps* policy first so people know how to deal with those.

SteadBytes: Great thank you! Sorry, I'm not sure what you mean by
> communication with upper apps policy

Could you please clarify? Did you mean define a way in which Flask-RESTPlus should interact with other applications?

Using `warnings/warn` is a common practice for Flask extensions that wish to communicate misconfiguration/possible error to the user. For example:
- [Flask-Login](https://github.com/maxcountryman/flask-login/blob/7e451269fbb7d58ad2530452a2d3420e0766ce3a/flask_login/login_manager.py#L100)
- [Flask-Debugtoolbar](https://github.com/mgood/flask-debugtoolbar/blob/d852042ccb68d0507e0ef8d9e0d53287ca766f3b/flask_debugtoolbar/__init__.py#L214)
- [Flask-Limiter](https://github.com/alisaifee/flask-limiter/blob/c0ae4d944bf233dccd68207c4de9e2c242cb055f/flask_limiter/extension.py#L112)

Or using `logger.warning` is also a common approach.

ziirish: > Could you please clarify? Did you mean define a way in which Flask-RESTPlus should interact with other applications?

I mean we should document what kind of *communication* people using FLask-RESTPlus should expect.
I know `warnings.warn` is a common practice, but we should document this.
The reason is the default filter might hide some messages hence this doesn't provide any help to the users:

```python
import warnings

warnings.warn('You should not use this', DeprecationWarning)
warnings.warn('Hey, something went wrong', UserWarning)
```

Result in:

```
$ python /tmp/warn.py
/tmp/warn.py:4: UserWarning: Hey, something went wrong
  warnings.warn('Hey, something went wrong', UserWarning)
```

So by default, unless people know they should expect a `DeprecationWarning` they won't notice it.

SteadBytes: Ah yes I understand - thank you for clarifying :smile: Yes I agree that this should definitely be documented. How about changing this:
https://github.com/noirbizarre/flask-restplus/pull/637/files#diff-dce4048f760a2254db20f3a78d7fcfa1R362 to say something like:

> Flask-RESTPlus will always fallback to the default ``json.dumps``  *serializer* if it cannot manage to import the one you configured. This is indicated by a ``warning`` during initialisation.

And then include an example of the output when starting the application, for example using `logging`:

```
* Serving Flask app "t" (lazy loading)
 * Environment: production
   WARNING: Do not use the development server in a production environment.
   Use a production WSGI server instead.
 * Debug mode: on
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
 * Restarting with stat
unable to import RESTPLUS_JSON_SERIALIZER=broken_seralizer.dumps, falling back to json.dumps
 * Debugger is active!
 * Debugger PIN: 627-628-298
```

Or using `warnings`:
```
 * Serving Flask app "t" (lazy loading)
 * Environment: production
   WARNING: Do not use the development server in a production environment.
   Use a production WSGI server instead.
 * Debug mode: on
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
 * Restarting with stat
/home/ben/Documents/flask-restplus/flask_restplus/api.py:213: UserWarning: unable to import RESTPLUS_JSON_SERIALIZER=broken_seralizer.dumps, falling back to json.dumps
  warnings.warn("unable to import RESTPLUS_JSON_SERIALIZER=broken_seralizer.dumps, falling back to json.dumps")
 * Debugger is active!
 * Debugger PIN: 627-628-298
```

ziirish: The only problem I see with this is the *code path* in which this warning will be raised is not deterministic (it will happen only after **some** *route* returns and if it needs marshalling).

Maybe the solution is to load the serializer within the `Api` constructor.

This way, people know when/where to catch the `warnings`.

ziirish: I have added the `warnings.warn` as suggested.
This should happen early in the initialization of the Api unless you pass it a Blueprint instead of a Flask app (in which case the warning will be raised when calling a route that needs serialization.
The loaded serializer is *cached* within the `app.config`. I don't think this is the best place to do so, but I didn't find any other...

PR #555, opened 2018-11-19
Fix $ref issue of not importing Json SchemaModels

tincumagic:
There are no more errors shown in Swagger Doc and the models sections in Swagger Doc show correctly the sub-schema in $ref fields. You will still need to register all models and sub-models that you expect on response or request. You register them with: api.schema_model(<NameOfModel>, <JSONSchema>)
This fix fixes the errors found in the the example found in the restplus documentation:
http://flask-restplus.readthedocs.io/en/stable/marshalling.html#define-model-using-json-schema

The fix related to the issue: JSON Schema $ref not working (not even the doc example) #275

Issue #414, opened 2018-03-28
Using nested JSON schemata for marshalling


Midnighter:
Hi,

In my application I would like to generate models for marshalling from JSON schema. The problem I run across is that flask-restplus does not seem to be aware of the whole JSONSchema spec. Specifically, I have the following schema:

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "title": "Plasmids Serialization Schema",
  "type": "object",
  "required": [
    "plasmids"
  ],
  "properties": {
    "plasmids": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/Plasmid"
      },
      "uniqueItems": true
    }
  },
  "additionalProperties": false
}
```

The error I get is then:

```
File "x/lib/python3.6/site-packages/flask/app.py", li
ne 1612, in full_dispatch_request                                                             
    rv = self.dispatch_request()                                                              
  File "x/lib/python3.6/site-packages/flask/app.py", line 1598, in dispatch_request                   
    return self.view_functions[rule.endpoint](**req.view_args)                                
  File "x/lib/python3.6/site-packages/flask_restplus/ap
i.py", line 313, in wrapper                                                                   
    resp = resource(*args, **kwargs)                                                          
  File "x/lib/python3.6/site-packages/flask/views.py", 
line 84, in view                                                                              
    return self.dispatch_request(*args, **kwargs)                                             
  File "x/lib/python3.6/site-packages/flask_restplus/re
source.py", line 44, in dispatch_request
    resp = meth(*args, **kwargs)
  File "x/lib/python3.6/site-packages/flask_restplus/ma
rshalling.py", line 110, in wrapper
    return marshal(resp, self.fields, self.envelope, mask)
  File "x/lib/python3.6/site-packages/flask_restplus/ma
rshalling.py", line 54, in marshal
    for k, v in fields.items())
AttributeError: 'SchemaModel' object has no attribute 'items'
```

Of course, my model has no attribute `'items'` since that is only the specification of the `"plasmids"` array. I have checked all my schemata against the `Draft4Validator` and they're fine. Am I using it in the wrong way within flask-restplus?

```python
with open_text(sanger_service.schemata, "plasmid.json",
               encoding="utf-8") as file_handle:
    plasmid = api.schema_model('Plasmid', json.load(file_handle))

with open_text(sanger_service.schemata, "plasmids.json",
               encoding="utf-8") as file_handle:
[...]

Issue #275, opened 2017-04-24
JSON Schema $ref not working (not even the doc example)


JenusL:
Using the JSON Schema example (http://flask-restplus.readthedocs.io/en/stable/marshalling.html#define-model-using-json-schema)

Address object looks like it's recognized but not parented correctly. This is on 0.10.1

![model](https://cloud.githubusercontent.com/assets/23527584/25333737/c20523f4-28eb-11e7-9cec-46969c7e14b3.png) ![example](https://cloud.githubusercontent.com/assets/23527584/25333738/c207cf46-28eb-11e7-93f7-4eebac39cd3c.png)

This is a stopper for me right now. Any workaround / ideas?

JenusL:
I haven't found a solution for this yet. Does anyone have a clue?

siddharthist:
What's the expected behavior?

JenusL:
Expected behavior would be that 'address' is showing in the Example box. It's completely absent right now.
It should also show the fields under Address in the Model box.
I have attached a very simple example where the same models are defined with JSON Schema and Fields.
They should be identical in swagger doc.
```
#! /usr/bin/env python
# -*- coding: utf-8 -*-

from flask import Flask
from flask_restplus import Api, Resource, fields

import json


app = Flask(__name__)
api = Api(app)


# Fields model is working as expected.
f_address = api.model('Address_', {
    'road': fields.String()
    })

f_person = api.model('Person_', {
    'name': fields.String(),
    'age': fields.Integer(),
    'birthday': fields.DateTime(),
    'address': fields.Nested(f_address)
    })


# The code below are directly from http://flask-restplus.readthedocs.io/en/stable/marshalling.html#define-model-using-json-schema
address = api.schema_model('Address', {
    'properties': {
        'road': {
            'type': 'string'
        },
    },
    'type': 'object'
})

person = address = api.schema_model('Person', {
    'required': ['address'],
    'properties': {
        'name': {
            'type': 'string'
        },
        'age': {
            'type': 'integer'
        },
        'birthdate': {
            'type': 'string',
            'format': 'date-time'
        },
        'address': {
            '$ref': '#/definitions/Address',
        }
    },
    'type': 'object'
})


# One route for each model
@api.route('/addressfields')
class AddressFields(Resource):
    @api.expect(f_person)
    def get(self):
        return 200

@api.route('/addressjson')
class AddressJson(Resource):
    @api.expect(person)
    def get(self):
        return 200


[...]

PR #640, opened 2019-05-07
Json schema include nested models in register_model

flayman:
When a SchemaModel type is created passing a Marshmallow schema into OpenAPIConverter.schema2jsonschema from the latest apispec.ext.marshmallow.openapi (and probably through other means), where a nested model is found it is referenced through a '$ref' key in the JSON with the value of "#/definitions/[submodel name]", where [submodel name] is the name given to the nested schema model. However, swagger does not add the nested schema model to the definitions, resulting in errors. This patch solves that problem by parsing out the referenced model name, finding that named model in api.models and registering that.

Tox and qa tasks are passing.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ Current
 - Fix `@api.expect(..., validate=False)` decorators for an :class:`Api` where `validate=True` is set on the constructor (:issue:`609`, :pr:`610`)
 - Ensure `basePath` is always a path
 - Hide Namespaces with all hidden Resources from Swagger documentation
+- In register_model() when working with SchemaModel type, also recursively register nested models
 
 0.12.1 (2018-09-28)
 -------------------
--- a/flask_restplus/swagger.py
+++ b/flask_restplus/swagger.py
@@ -16,7 +16,7 @@ from flask import current_app
 from werkzeug.routing import parse_rule
 
 from . import fields
-from .model import Model, ModelBase
+from .model import SchemaModel, Model, ModelBase
 from .reqparse import RequestParser
 from .utils import merge, not_none, not_none_sorted
 from ._http import HTTPStatus
@@ -557,6 +557,12 @@ class Swagger(object):
         if isinstance(specs, Model):
             for field in itervalues(specs):
                 self.register_field(field)
+        if isinstance(specs, SchemaModel):
+            for prop in itervalues(specs._schema['properties']):
+                if ('$ref' in prop):
+                    name = prop['$ref'].rsplit('/', 1).pop()
+                    if name in self.api.models:
+                        self.register_model(self.api.models[name])
         return ref(model)
 
     def register_field(self, field):


j5awry:
@flayman thank you for your contribution. Would you be able to do the following to help expedite?

1. please add unit tests for the fix
2. Could you make sure this matches an issue? There are a bunch of related issues, and we want to make sure we can match PRs to Issues: #275 #414 are possible matches

I'm going to do some more reading on Marshmallow as well to make sure i'm 100% up on how its working.

flayman:
Hi John,

This represents an edge case that may not fall squarely within the use of
flask-restplus. I had also submitted a pull request to the apispec project
because I was using their plugin incorrectly and thought I needed to use a
non-trivial schema name resolver, but it turns out the best course of
action is to not resolve nested schema names as this causes the nested
schema to be included inline instead of by reference. I don't think they
are going to merge my changes:
https://github.com/marshmallow-code/apispec/pull/447

I also don't know whether flask-restplus will encounter the error I found
in normal use cases, but it probably doesn't hurt to handle it. I'll have a
look at those two issues and see about test cases.

Regards,
Matt

flayman:
Hi again. I've had a look at those two issues and they seem to be the same
thing. They also seem to show that this is a real use case. Issue #275
offers a solution in the form of a patch to swagger.py, just like my pull
request, but mine is a lot simpler. It's just six additional lines of code
and one change to the imports. There's no need for a new method for
registering json refs. I'll see about tests. I've only been interested on
one level of nesting. Ideally the solution would handle multiple levels and
deal with circular references.

Regards,
Matt

j5awry:
PR #555 doesn't work as advertised, especially after a rebase. That's another reason why I'm keen on unittests.

There's a core issue of jsonschema, and anything using $ref, being of unknown depth and complexity, and needing to resolve refs which _may_ be circular (which is allowed in jsonschema). I've suggested using the `jsonref` library which appears to work in most cases. However, i'm still evaluating more things. 

I'll give this one a try too and see if it solves complex cases.

flayman:
I'm a little bit busy just now with something else and my own requirements are mostly satisfied, but as soon as I have a bit of time I'll get on with some test cases. Hopefully in the next few days.

flayman:
Hi. I tried the example you gave in PR #555 and it does not crash on the KeyError when run with my branch installed. (Actually, in order to get the example to run I needed to add an import for Namespace and a line with "ns = api.namespace('test')". The reason for the hard error in the other patch is that it assumes the key will exist on the dictionary self.api.models. It needs to be more defensive.

However, as in that codebase mine also fails to include the nested model on the Api instance. The Swagger UI throws an error when you try to expand TEST_SCHEMA or view the endpoint, so the code is not doing enough. I'd seen this error in my own use case and the changes were sufficient to make that go away. I think that was because of the change that I'd asked apispec to merge, which would have handled it inside the OpenAPIConverter class. But this has nothing at all to do with Marshmallow, so I'll need to see what else is needed here.

flayman:
I haven't written the test case yet but it shouldn't be too difficult. I have pushed a change that works with the example. This is a simpler and more reliable approach.

j5awry:
Thanks! I'm heading out on a short vacation, so won't get a chance to check this, and even more complicated tests, till after I get back on Wednesday.

j5awry:
For anyone jumping in, i have a quick functional test file we can use. A couple simple cases are at the top:

https://gist.github.com/j5awry/aa8f78791078ddd63a3957ca584e5e60

flayman:
Okay, I have a question about how to write the test(s). This needs to test the result of calling register_model() in swagger.py, but nothing else tests that directly. The test_model.py test cases are not useful because the api is not working on them. test_namespace.py seems like the right place to do this, but all the models presented there are trivial. I would like to use your functional test schemas in there but it becomes entangled with the functionality of swagger.py so it may not be cleanly separated enough for unit testing. Do you have any suggestions?

flayman:
I've been reading the json schema specification and I think supporting references in flask-restplus is probably more complicated than attempts I've seen so far. json schemas can refer to internal definitions like...

`{ "$ref": "#/definitions/address" }`

 ... and also definitions in other json schemas as in...

`{ "$ref": "definitions.json#/address" }`
[https://json-schema.org/understanding-json-schema/structuring.html](https://json-schema.org/understanding-json-schema/structuring.html)

This is probably how it should be done in Swagger rather than bringing everything into Swagger's definitions and risking conflicts, but it could prove tricky. So for example, if a schema named Person brings in the definition "address", which it is referring to internally, other schemas could refer to it as "Person#/definitions/address". When nested models are registered, they would have to refer to the nested models in that form. This probably represents a major change.

j5awry:
@flayman you're correct about all the jsonschema stuff. references are a big headache. I've worked out the following test cases for jsonschemas so far:

jsonschema test cases

1) simple -- one layer, one key
2) flat -- one layer, many key
3) nested -- multiple layers, one key per
    3a) two layers, one key per
    3b) 10 layers, one key per
4) refed simple -- two layers, one ref. abstract 3a into refs
5) refed nested -- abstract 3b into refs

Right now with your changes we've got 1, 2, 4 down happily. I'm writing the tests for 3 now. I've used an example of a highly complex schema from work for 5, and it's failing. However, I want to simplify it down a bit more so I can track down where everything is happening.

In parallel to this PR, I'm going to further investigate using jsonref to de-reference things:

https://github.com/gazpachoking/jsonref

Since someone has solved the reference part, perhaps that's the best option for us to take, since it's not an easy problem.

flayman:
Hi John,

I can't comment on jsonref because I haven't looked at it. It may solve all or most of the problems. But I think that the reference problem is more complicated than you've indicated in your test cases, because there's no reason you can't have one schema reference a definition in another schema, and that creates all sorts of problems. The Swagger document has its own definitions, with each of the referenced models in there. The problem is that if you parse a json schema into a SchemaModel and that jsonschema has its own definitions, it's not clear where those should be placed. Do you put them into the Swagger definitions as global definitions or do you leave them where they are?

If you put the definitions onto the Swagger document's definitions property then there's a risk of collisions, and I think it also goes against the jsonschema spec. If you leave them where they are, then the model created from that json schema (perhaps called MyModel) won't be able to refer to its own definitions through Swagger, because in the Swagger document they will be found under #/definitions/MyModel/definitions. So when you register the model you'd have to actually change all the refs in the MyModel schema to point there.  That touches models.py. I actually did some work along these links which I haven't pushed to GitHub. It's a bit messy because a model with internal refs will show properties looking something like this:

TEST_SCHEMA {
    secret* [
        TEST_SCHEMA/definitions/key {
            key* | string
        }
    ]
}

It's not working correctly yet because the payload schema for expect is not finding the nested definition. I stopped work on it because it's a fairly radical change that the maintainers may not want to bless.

alvarolopez:
Hi.

Any idea when this PR would get merged? This is an issue we are facing right now, since we are using JSON schemas with nested models and they are not properly rendered.

flayman:
It's doubtful that that this PR will be merged. There were some fundamental questions about how to resolve external JSON refs. I think it has been overtaken.

PR #648, opened 2019-05-23
Add support for custom default model classes

### LINKED ISSUE #647: Feature Request: Override the default Model/OrderedModel class to use in .model()
Hello!

In order to get a little more flexibility on the api and to keep the current api.model() syntax that i like.  It would be great to be able to provide custom Model/OrderedModel classes to use.

One use-case for this would be to provide a model class that would be strict by default, automatically adding additionalProperties = False to the schema without having to modify how the whole api declares the models.

A pull request will be on the way soon.

Thank you for your great work!

lindycoder:
By specifying default_model_class and/or default_ordered_model_class in
the Api or the Namespace, the according class will be used when using
the api/namespace.model() method.

resolves #647

### DIFF
--- a/tests/test_namespace.py
+++ b/tests/test_namespace.py
@@ -56,6 +56,22 @@ class NamespaceTest(object):
         assert 'Person' in api.models
         assert isinstance(api.models['Person'], OrderedModel)
 
+    def test_custom_model(self):
+        class CustomModel(Model):
+            pass
+        api = Namespace('test', default_model_class=CustomModel)
+        api.model('Person', {})
+        assert 'Person' in api.models
+        assert isinstance(api.models['Person'], CustomModel)
+
+    def test_custom_ordered_model(self):
+        class CustomOrderedModel(Model):
+            pass
+        api = Namespace('test', ordered=True, default_ordered_model_class=CustomOrderedModel)
+        api.model('Person', {})
+        assert 'Person' in api.models
+        assert isinstance(api.models['Person'], CustomOrderedModel)
+
     def test_schema_model(self):
         api = Namespace('test')
         api.schema_model('Person', {})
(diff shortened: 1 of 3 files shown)


tests/test_namespace.py
```diff
@@ -56,6 +56,22 @@ def test_ordered_model(self):
...
 
+    def test_custom_model(self):
+        class CustomModel(Model):
+            pass
+        api = Namespace('test', default_model_class=CustomModel)
+        api.model('Person', {})
+        assert 'Person' in api.models
+        assert isinstance(api.models['Person'], CustomModel)
+
+    def test_custome_ordered_model(self):
```
SteadBytes: Little typo : `custome` -> `custom`  :smile:

tests/test_namespace.py
```diff
@@ -133,3 +149,22 @@ def post(self):
...
+
+    def test_custom_model(self, app):
+        class CustomModel(Model):
+            pass
+        api = restplus.Api(app, default_model_class=CustomModel)
+        api.model('Person', {})
+        assert 'Person' in api.models
+        assert isinstance(api.models['Person'], CustomModel)
+
+    def test_custome_ordered_model(self, app):
```
SteadBytes: Same type of before :smile:  `custome `->`custom`

SteadBytes:
Thanks for the PR - it's definitely a useful idea to be able to set these defaults, however I'm not sure this is the correct approach. For example, passing a default model class to the `Api` does not enforce that default on any `Namespace`'s attached to it which would lead to inconsistencies. However, this is further complicated by the fact that a `Namespace` can belong to multiple `Api`s, which could each be initialised with different defaults. 

I am also hesitant to keep adding more and more `kwarg`s to the `Api` and `Namespace` `__init__` methods as it's becoming difficult to manage and to understand what they all do. Perhaps exposing the default models as class properties and then allowing users to subclass the `Api` and `Namespace` classes to set a custom value may be a possible approach.

What do you think?

lindycoder:
I think that namespaces should stay as independent as they can be.  For very simple uses cases I really like that you can just make an Api object and start using it without caring about the namespacing and its advanced usage.  

The only reason I suggested to put in the Api constructor is to keep this simplicity and ideally all parameters a namespace could take might be passed there.   I see you currently don't use the **kwargs, maybe these could be integrally passed down to the namespace and have a clear entry in the doc : "Extra params passed to the default namespace" with a ref.

Have you though about the Api class extending the Namespace? Not sure if that's a good idea but an API current have all the methods of the namespace via the getattr bridging.

SteadBytes:
I agree that namespaces should be as independent as possible, however it seems to me that setting a default `Model` class at the `Api` level suggests that it should be applied application wide. It should then be *overridden* on a per namespace basis (similar to how disabling/enabling validation or Swagger documentation currently works for example) if desired. 

The Api constructor does make sense, as it's the logical place to configure options for the Api. My only concern is that there are already many `kwargs` going into that constructor and it may become difficult to manage.

I haven't directly considered that approach so I can't comment on it, however I will have a look and assess whether it may be useful. I will ask the other maintainers about all these points (Api vs Namespace, kwargs etc) as I definitely like this feature, but it would be beneficial to consider the pros and cons of how to *implement* it :smile:

lindycoder:
Not personally looking forward to this anymore.

PR #649, opened 2019-05-25
Documentation: the parameter should be called readonly

aicioara:
Problem is here
https://flask-restplus.readthedocs.io/en/stable/example.html

As per documentation here 

https://flask-restplus.readthedocs.io/en/stable/api.html#flask_restplus.fields.Raw

The field is called `readonly`, not `readOnly`.

This fails silently, leading to long sessions of debugging.


SteadBytes:
Thanks for this PR - well spotted! I will merge this now :+1:

SteadBytes:
Thanks for this PR - well spotted! I will merge this now :+1:

PR #654, opened 2019-06-14
Fix @app.param and @app.header producing invalid  Swagger/OpenAPI document

Drarok:
Only include description in param and header when it has a value, else the resulting document fails validation

### DIFF
--- a/flask_restplus/namespace.py
+++ b/flask_restplus/namespace.py
@@ -267,9 +267,10 @@ class Namespace(object):
         :param str description: a small description
         :param str _in: the parameter location `(query|header|formData|body|cookie)`
         '''
-        param = kwargs
+        param = kwargs.copy()
         param['in'] = _in
-        param['description'] = description
+        if description is not None:
+            param['description'] = description
         return self.doc(params={name: param})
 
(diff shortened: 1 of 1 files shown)


j5awry:
Can you provide the open issue this is in relation to? And/or recommended test cases?

Also, we'd greatly appreciate if unit tests could be added. If you're not sure where, we can help guide on that. Minimally we need steps to reproduce the error so we can test the fix.

Drarok:
There's no open issue at present, it's something I noticed was incorrect when trying to validate the `swagger.json` for our API.

I'll have a look at producing a test case, but unless there's already something validating against the Swagger schema, it might be a lot of work. I can assert that description isn't null (which is what the problem is) directly, but I don't know if that's official enough?

That said, I've found the [contributing docs] and they appear to be incorrect:

```bash
pip install -e .[dev]
zsh: no matches found: .[dev]
```

[contributing docs]: https://flask-restplus.readthedocs.io/en/latest/contributing.html

Drarok:
To put it briefly, the `description` key must be a string, so `null` is invalid: https://swagger.io/specification/#parameterObject

Issue #288, opened 2017-05-18
Duplicate Swagger operationIds when using multiple routes for same resource


jwmullally:
Hi, (this is a minor issue, so no sweat)

When multiple routes are used for the same resource, the Swagger code generates the same operationId for each resource. This violates the Swagger spec and confuses some clients (see below).           

Is there an existing good way to handle this?

Thanks for the great library guys:)

jwmullally:
### Swagger validation and client generation issues from duplicate operationId

Test code:
(Based on https://github.com/postrational/rest_api_demo/blob/14432600d83979153a50a2d0fa80c4ab8c0a844b/rest_api_demo/api/blog/endpoints/posts.py#L73 ). (Its debatable whether this many-route -> one-resource pattern should be allowed versus changing this to a single paramaterised POST query, but other libraries conclude it should be supported, e.g. https://github.com/flask-restful/flask-restful/issues/114 )

```python
#!/usr/bin/env python
from flask import Flask, request
from flask_restplus import Api, Resource, fields

app = Flask(__name__)
api = Api(app)

archive = api.model('Archive', {
    'ret_year': fields.Integer(readOnly=True),
    'ret_month': fields.Integer(required=True),
    'ret_day': fields.Integer(required=True)
})

@api.route('/archive/<int:year>/')
@api.route('/archive/<int:year>/<int:month>/')
@api.route('/archive/<int:year>/<int:month>/<int:day>/')
class ArchiveResource(Resource):
    @api.marshal_with(archive)
    def get(self, year, month=-1, day=-1):
        """
        Returns list of blog posts from a specified time period.
        """
        return {'ret_year': year, 'ret_month': month, 'ret_day': day}

if __name__ == "__main__":
    app.run(debug=True)
```
Output: [swagger.json.txt](https://github.com/noirbizarre/flask-restplus/files/1009860/swagger.json.txt)

Validation with http://editor.swagger.io :
```
Errors

Semantic error at paths./archive/{year}/{month}/.get.operationId
operationIds must be unique
Semantic error at paths./archive/{year}/{month}/{day}/.get.operationId
operationIds must be unique
```

**bravado** client usage
```python
In [1]: from bravado.client import SwaggerClient
In [2]: client = SwaggerClient.from_url('http://localhost:5000/swagger.json')
In [3]: dir(client.default)
Out[3]: ['get_archive_resource']
In [4]: client.default.get_archive_resource.operation.path_name
[...]

PR #658, opened 2019-06-26
Multiple Routes Swagger Documentation 

SteadBytes:
# Motivation
See #288 for initial issue. 

Using the `doc` parameter of `Api.route()` decorator applies documentation to the `Resource` as a whole. This prevents documentation being added on a per-route basis (for example adding an additional description to a route, or marking one route as deprecated) as the `doc` is applied to all routes.

Furthermore, each route uses the same Swagger `operationId`, causing both routes to be expanded at the same time when selecting one in the Swagger UI (the elements are selected using an HTML `id` which is set to the `operationId`.

# Changes
- Using the `doc` parameter `Api.route()` will apply the documentation to that route only
- Routes using the `doc` parameter are given unique `operationId`s (combining the original `operationId` with the URLs provided)
- Each route will inherit documentation from the `Resource` (i.e. if applied using `@api.doc`, unless explicitly overridden with the `doc` parameter

# Examples

## Providing a `description` to a single route:

```python
@api.route("/my-resource/<id>")
@api.route(
    "/also-my-resource/<id>",
    doc={"description": "Alias for /my-resource/<id>"},
)

class MyResource(Resource):
    def get(self, id):
        return {}
```
single_route_1

Expanding the first route only (note the `description`):
single_route_2

Expanding the second route only (note lack of `description`):
single_route_3

## Inheriting/overriding from the `Resource` 
```python
@api.route('/my-resource/<id>', '/test/<id>')
@api.route('/also-my-resource/<id>', doc={"description": "Alias for /my-resource/<id>"})
@api.doc(params={'id': 'An ID'}, description="My resource")
class MyResource(Resource):
    def get(self, id):
        return {}
```
image

Expanding the first route only (note `id` parameter from ``@api.doc()`` but overridden `description`)
image

Expanding the second route only (note `id` parameter and `description` inherited from `@api.doc()` decorator):
image

Various combinations of the above can be used with of the available documentation parameters (`description`, `id`, `params`, `deprecated` etc).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ Current
 - Fix `@api.expect(..., validate=False)` decorators for an :class:`Api` where `validate=True` is set on the constructor (:issue:`609`, :pr:`610`)
 - Ensure `basePath` is always a path
 - Hide Namespaces with all hidden Resources from Swagger documentation
+- Per route Swagger documentation for multiple routes on a ``Resource``
 
 0.12.1 (2018-09-28)
 -------------------
--- a/doc/swagger.rst
+++ b/doc/swagger.rst
@@ -355,6 +355,63 @@ For example, these two declarations are equivalent:
         def get(self, id):
             return {}
 
+Multiple Routes per Resource
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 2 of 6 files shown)


copdips:
even it's late, but a big thank to you.

PR #669, opened 2019-07-12
Feature: strict string field enum

Drarok:
This PR adds a `strict_enum` option to `String` field, which raises a `MarshallingError` if the supplied value is not in the `enum` option.


Drarok:
Looks like this is already handled for HTTP requests, just not at a lower level.

PR #673, opened 2019-07-19
Support envelope parameter in Swagger documentation #390

### LINKED ISSUE #390: marshal_with envelope does not affect Swagger documentation
Consider the following code:
```python
from flask_restplus import Api, fields
from flask import Blueprint, request

api_example = Blueprint("api_example", __name__)
api = Api(api_example, description="Example API")

input_fields = api.model('InputFieldsModel', {
    'file': fields.String(description='File name', required=True),
    'number': fields.Integer(min=1, default=1),
})

input_model = api.model("InputModel", {
    "files": fields.List(fields.Nested(input_fields), required=True)
})

output_model_fields = api.model("OutputModelFields", {
    "file": fields.String(required=True),
    "data": fields.Float(required=True)
})

@api.route("/example")
class Example(Resource):
    @api.expect(input_model, validate=True)
    @api.marshal_with(output_model_fields, envelope="results")
    def post(self):
        return [{"file": "abc", "data": 0.1}, {"file": "xyz", "data": 0.9}]
```

I think that expected value in Swagger Documentation for the `POST` endpoint should have the following form:
*Model Example Value*:
```json
{"result": [{"file": "string", "data": 0}, {"file": "string", "data": 0}]}
```

But it actually is:
```json
{"file": "string", "data": 0}
```

which means that automated documentation generation **does not consider** `envelope` parameter, which is misleading when looking at Swagger UI.
Is it a bug or is there another way of generating proper documentation based on my requirements.

tnt-dev:
Fixes #390 

If `envelope` presents in `marshal_with` options response schema should have the following form:
```
{"properties": {envolope_key: response_model}}
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Current
 - Ensure `basePath` is always a path
 - Hide Namespaces with all hidden Resources from Swagger documentation
 - Per route Swagger documentation for multiple routes on a ``Resource``
+- Support ``envelope`` parameter in Swagger documentation (:pr:`390`)
 
 0.12.1 (2018-09-28)
 -------------------
--- a/flask_restplus/namespace.py
+++ b/flask_restplus/namespace.py
@@ -238,7 +238,7 @@ class Namespace(object):
         def wrapper(func):
             doc = {
                 'responses': {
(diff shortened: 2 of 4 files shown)


j5awry:
TY for the rebase. Will do one last check with the team, but otherwise looks good.

Issue #685, opened 2019-08-15
Schemas added to namespace.models are dropped


erinpentecost:
When adding schemas to a namespace instance's models collection, they are dropped from the generated swagger spec if they are not used by API endpoints through the use of marshal or expects decorators. This change forces those to be included.

I was running into issues with nested definitions being dropped.

j5awry:
> I was running into issues with nested definitions being dropped.

Just for clarification, were the nested schemas included in schemas that are being used?

brief pseudo:

```
endpoint A uses Schema A
Schema A has nested Schema B
Schema B is not rendered in OpenAPI
```

erinpentecost:
Yes, that's right. I was registering them with schema_model. The parent schema owns a union of child schemas, and the child schemas were not being included.

PR #686, opened 2019-08-15
Include all models in Swagger

erinpentecost:
Issue #685 .

### DIFF
--- a/flask_restplus/swagger.py
+++ b/flask_restplus/swagger.py
@@ -381,6 +381,8 @@ class Swagger(object):
         doc = self.extract_resource_doc(resource, url, route_doc=route_doc)
         if doc is False:
             return
+        for _n, _v in ns.models.items():
+            self.register_model(_v)
         path = {
             'parameters': self.parameters_for(doc) or None
         }
--- a/tests/test_swagger.py
+++ b/tests/test_swagger.py
@@ -2166,7 +2166,7 @@ class SwaggerTest(object):
         data = client.get_specs()
 
         assert 'definitions' in data
-        assert 'Person' not in data['definitions']
+        assert 'Person' in data['definitions']
         assert 'Child' in data['definitions']
 
         path = data['paths']['/extend/']


j5awry:
Thank you for contributing. This change caused a test failure. Please look into the test failure, and ensure all tests pass locally. 

```
SwaggerTest.test_clone 
self = <tests.test_swagger.SwaggerTest object at 0x7f636e2aca90>
api = <flask_restplus.api.Api object at 0x7f636d7d2400>
client = <TestClient <Flask 'tests.conftest'>>
    def test_clone(self, api, client):
        parent = api.model('Person', {
            'name': restplus.fields.String,
            'age': restplus.fields.Integer,
            'birthdate': restplus.fields.DateTime,
        })
    
        child = api.clone('Child', parent, {
            'extra': restplus.fields.String,
        })
    
        @api.route('/extend/')
        class ModelAsDict(restplus.Resource):
            @api.doc(model=child)
            def get(self):
                return {}
    
            @api.doc(model='Child')
            def post(self):
                return {}
    
        data = client.get_specs()
    
        assert 'definitions' in data
>       assert 'Person' not in data['definitions']
E       AssertionError: assert 'Person' not in {'Child': {'properties': {'age': {'type': 'integer'}, 'birthdate': {'format': 'date-time', 'type': 'string'}, 'extra':...e': 'integer'}, 'birthdate': {'format': 'date-time', 'type': 'string'}, 'name': {'type': 'string'}}, 'type': 'object'}}
tests/test_swagger.py:2069: AssertionError
```

erinpentecost:
Thank you for the test results! This change intentionally adds any models declared on a namespace, even if they are not used. The failing test makes sure that unused models aren't included. I'm not sure why that test case exists, but the fix (for this merge request) would be to remove/invert that portion of the test.

kdhlab:
I was having the same issue with nested models that were not explicitly called by a method, and applying this change worked for me. Can we get this pulled in?

j5awry:
Please see the above comment -- there's a failing test. The core functionality of the test is inverted in this PR. We request that the creator of PRs ensure tests are passing, and update tests as required.

erinpentecost:
The unit test explicitly checks for the existing behavior, and I'm not sure what the reasoning was for that test. If it's not an important test, then fine, I can delete it. But I don't know what impact the change is going to have on the framework as a whole.

j5awry:
@erinpentecost, correct, the test specifically checks for the current behaviour. This changes the behaviour. What needs to happen is this test either needs altered to reflect this PRs desired behaviour, or this test should be deleted and a _new_ test added to test the new behaviour.

Judging by the test, the "intended" behaviour is to ignore unused models. So part of this is a proposal to change the old behaviour. But based on our [contribution doc ](https://github.com/noirbizarre/flask-restplus/blob/master/CONTRIBUTING.rst) 

> your contribution should be tested and the test suite should pass successfully

So for the maintainers to review it we have to ensure the functionality in the unit tests. From there, maintainers can then decide on the approach. Considering this is a fairly large change in functionality, I'm guessing there'll be a bit of debate (I'll open a little of that now so we can start discussing while tests are getting updated).

flask_restplus/swagger.py
```diff
@@ -377,6 +377,8 @@ def serialize_resource(self, ns, resource, url, route_doc=None, **kwargs):
         doc = self.extract_resource_doc(resource, url, route_doc=route_doc)
         if doc is False:
             return
+        for _n,_v in ns.models.items():
```
j5awry: This would include all models. We might want to consider adding a switch for users to have either the new functionality or the old functionality.

j5awry:
Discussion:

To restate, this change would include all models not just those in use. Some consideration points:

1. Should we add a switch so users can choose the functionality?
2. What sort of change management should we do on this change? It's small in code but large in implication. Things like "in progress" work may show up that users were used to having hidden.

Those are beyond the general case of "should we do this?" Based on testing and code, it appears the intent of the author was to hide unused things. However, if a large enough portion of use want all available, it's worth adding. 

I'd vote for adding in a switch, and doing a more robust release guide on the version. I'd also suggest the default be to _not_ show the models, so current users are not surprised. opt-in vs. opt-out.

PR #719, opened 2019-09-30
Fixing Swagger Issue when using @api.expect() on a request parser

tnweiss:
Below is a request parser for a simple hello world server.

```python
parser = reqparse.RequestParser()
parser.add_argument('test', type=int, location='json')
parser.add_argument('test1', location='json')
```

The resulting swagger JSON for these parameters is 

```json
"parameters": [
    {
         "name": "test",
          "in": "body",
          "type": "integer"
     },
     {
          "name": "test1",
           "in": "body",
           "type": "string"
       }
]
```

The result should be something similar to when a user provides a model with api.expect()

```json
"parameters": [
    {
        "name": "payload",
        "required": true,
        "in": "body",
        "schema": {
            "properties": {
                "test": {
                    "type": "integer"
                },
                "test1": {
                    "type": "string"
                }
            }
        }
    }
]
```

These changes fix that bug.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,6 @@ doc/_build/
 # Specifics
 flask_restplus/static
 node_modules
+
+# PyCharm
+.idea*
--- a/flask_restplus/swagger.py
+++ b/flask_restplus/swagger.py
@@ -143,6 +143,31 @@ def is_hidden(resource, route_doc=None):
         return hasattr(resource, "__apidoc__") and resource.__apidoc__ is False
 
 
+def rparser_to_swagger_body_param(request_parser):
+    """
+    If any parameters are in the request body then return the swagger representation for the requests json body
+    :param request_parser: The request parser containing params for the request
+    :return:
+    """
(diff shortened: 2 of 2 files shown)


wood-j:
Great job, Thanks
While it's not avaliable at 2020/11, I just use your commit as a monkey patch so solve my problem.
Thanks aloooooooooooooooot!

By the way, I found that `'type': 'object'` should be inside of `schema` for version `0.13.0`

```python
def rparser_to_swagger_body_param(request_parser):
    """
    If any parameters are in the request body then return the swagger representation for the requests json body
    :param request_parser: The request parser containing params for the request
    :return:
    """
    json_body_list = [p for p in request_parser.__schema__ if p['in'] == 'body']
    if not json_body_list:
        return

    properties = {}
    for param in json_body_list:
        properties[param['name']] = {
            'type': 'string' if 'type' not in param else param['type']
        }

    return {
        'name': 'payload',
        'required': True,
        'in': 'body',
        'schema': {
            'type': 'object',
            'properties': properties
        }
    }
```

PR #722, opened 2019-10-02
Fix passing of kwargs to `init_app`

jethroguce:
initiating app instance with arguments does not take effect,
```
api = Api(app, add_specs=False)
```

but when using the lazy instantiation it works

```
api = Api()
api.init_app(app, add_specs=False)
```

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -149,7 +149,7 @@ class Api(object):
 
         if app is not None:
             self.app = app
-            self.init_app(app)
+            self.init_app(app, **kwargs)
         # super(Api, self).__init__(app, **kwargs)
 
     def init_app(self, app, **kwargs):


jethroguce:
@j5awry can you have a quick look at this

j5awry:
1. we should check the tests for coverage of this new call
2. looking at the history, git blame, and the commented portion, it looks like this capability existed before. Looks like 4 years ago there was a switch, and the old way did pass kwargs to a parent class (that looks like it _did_ do what this is asking for)

From that standpoint, it's worth asking if this is _meant_ to, since it seems to be a conscious change. I'm in favor of passing kwargs down though.

SteadBytes:
I'm also in favour of passing kwargs down, however it would certainly be useful to understand the original reason for removing it if possible. 

2 things before merging if you don't mind:
1. Cover the change with tests
2. Remove https://github.com/noirbizarre/flask-restplus/pull/722/files#diff-31928362e95b5b97b24d2ccf01506003R153 - may as well clean it up a bit whilst here :smile:

PR #732, opened 2019-10-14
Import the ABCs from 'collections.abc' instead of 'collections'

### LINKED ISSUE #727: Python 3.8 support
Using latest flask-restplus version with Python 3.7.4, I have the following warning:

```
flask_restplus/model.py:8: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working

    from collections import OrderedDict, MutableMapping
```

It is not fixed on master branch and I cannot find any open item with "DeprecationWarning" matching this.

I can submit a PR to solve this if that's ok?

Colin-b:
Import the ABCs from 'collections.abc' instead of 'collections' by default as it is deprecated since Python3.7, and in 3.8 it will stop working.

I considered that Python3 is the de-facto standard as Python2 support is being dropped in every other Python project and support ends at end of year. This is why I removed the OrderedDict sample and use a builtin dict instead of they are ordered since 3.6

I however kept Python2 support in the unlikely event that some projects are still upgrading flask-restplus dependency while not upgrading to Python3.

I did not add any additional test case as I assume the test suite runs with Python2 as well as 3.X, if this is not the case I am more than happy to remove the extra TODO and try/except blocks in this PR.

closes #727

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ Current
 -------
 
 - Ensure that exceptions raised in error handler, including programming errors, are logged (:issue:`705`, :pr:`706`)
+- Import the ABCs from 'collections.abc' instead of 'collections' by default as it is deprecated since Python3.7, and in 3.8 it will stop working. Python2.7 is still supported though.
 
 0.13.0 (2019-08-12)
 -------------------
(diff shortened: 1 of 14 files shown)


andreykurilin:
looks good to me

ziirish:
Thanks

PR #738, opened 2019-10-17
Update deprecated import for werkzeug's cached_property

jonblum:
Resolves `DeprecationWarning: The import 'werkzeug.cached_property' is deprecated and will be removed in Werkzeug 1.0. Use 'from werkzeug.utils import cached_property' instead.` raised by Werkzeug 0.19+.

`cached_property` has always resided in werkzeug.utils, so this is backwards-compatible.

### DIFF
--- a/flask_restplus/fields.py
+++ b/flask_restplus/fields.py
@@ -14,7 +14,7 @@ from six import iteritems, itervalues, text_type, string_types
 from six.moves.urllib.parse import urlparse, urlunparse
 
 from flask import url_for, request
-from werkzeug import cached_property
+from werkzeug.utils import cached_property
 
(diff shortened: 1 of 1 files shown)


andreykurilin:
looks good to me and should be compatible with minimum required versions of packages:

Flask >=0.8 https://github.com/noirbizarre/flask-restplus/blob/9b9ad423a80f51d374ee6d4e40d164b8f9844cfb/requirements/install.pip#L3
Werkzeug>=0.6.1 https://github.com/pallets/flask/blob/0.8/setup.py#L93
werkzeug.utils.cached_propert exists at https://github.com/pallets/werkzeug/blob/0.6.1/werkzeug/utils.py#L30

ziirish:
Thanks

PR #741, opened 2019-10-21
Don't return exception when errorhandler is set (fix #693)

### LINKED ISSUE #693: Latest 0.13.0 release breaks errorhandler()
My code that overrides an exception via errorhandler() that has worked forever broke with the latest 0.13.0 release. Falling back to 0.12.1 fixes the problem. Since it has been so long since there's been a release and the changelog is very hard to read, I have not been able to track down the code change that produced this problem yet.

```python
@api.errorhandler(SomeExistingError)
def handle_existing_exception(error):
    """Some Existing failure"""
    return dict(message=f"{error.error}: {error.description}"), error.status_code
```

brmzkw:
If errorhandler is configured, error should not be returned to client
even if PROPAGATE_EXCEPTIONS is true.

### DIFF
--- a/flask_restplus/api.py
+++ b/flask_restplus/api.py
@@ -593,7 +593,12 @@ class Api(object):
         '''
         got_request_exception.send(current_app._get_current_object(), exception=e)
 
-        if not isinstance(e, HTTPException) and current_app.propagate_exceptions:
+        # When propagate_exceptions is set, do not return the exception to the
+        # client if a handler is configured for the exception.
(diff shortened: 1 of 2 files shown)


brmzkw:
@SteadBytes I added the comments you wanted.

SteadBytes:
Great thank you for your contribution!

brmzkw:
Thanks for the merge. Do you know when the next release is planned?

PR #742, opened 2019-10-22
Skip choices for nullable argument in RequestParser

adzhurinskij:
Hi.

It's fix popular case with arguments which can be choices or nullable.

```python
parser = reqparse.RequestParser()
parser.add_argument('description', type=str, store_missing=False, nullable=True)
args = parser.parse_args(strict=True)
```


andreixk:
@adzhurinskij You need to modify unit tests accordingly in order to ensure the checks are passing.

Issue #716, opened 2019-09-20
How can I write a custom validator?


AbdoDabbas:
Is there a way to write a validator for a custom field type that I can use in the API model (in except decorator).
I mean like this:
```
api.model('a_name', {
    'email': MyEmailType(required=True)
})
```

I tried to inherit from `fields.String` and did it like:
EMAIL_REGEX = re.compile(r'\S+@\S+\.\S+')
```
class Email(fields.String):
    """
    Email field
    """
    __schema_type__ = 'string'
    __schema_format__ = 'email'
    __schema_example__ = 'email@domain.com'
 
    def validate(self, value):
        if not value:
            return False if self.required else True
        if not EMAIL_REGEX.match(value):
            return False
        return True
```

AbdoDabbas:
I tried to apply this with no luck:
https://aviaryan.com/blog/gsoc/restplus-validation-custom-fields

JBarrioGarcia:
I don't know how to do a custom validator, but you can use "pattern" to validate the string:

api.model('a_name', {
    'email': fields.String(required=True, pattern='\S+@\S+\.\S+')
})

I hope it helps

j5awry:
@AbdoDabbas , reading what you implemented, and your described problem, did you implement the last step? In the example you posted, they created another function, `validate_payload()`. While the description is incomplete, I inferred that this method was then called on each endpoint after serializing the payload. something like:

```
@route("/test")
class MyResource(Resource):
     @api.expect(MY_MODEL, validate=False)
     def post(self):
          payload = api.payload
          validate_payload(payload, MY_MODEL)
```

This is based on the code in the flask_restplus `Resource` class: https://github.com/noirbizarre/flask-restplus/blob/master/flask_restplus/resource.py#L75

I'm not having good luck on peeking where `validate_payload` is called though.

AbdoDabbas:
@JBarrioGarcia I tried your way, the problem now is the error message is strange, it says:
```
'string' does not match '\\\\S+@\\\\S+.\\\\S+'
```
If 'string' can be resolved using the "title", I didn't find how to replace the pattern in the message to be something I want.
So the idea is, is there a way to replace the error message?

andreixk:
+1. the `pattern=...` solution works, but it gives a really ugly message and confuses users
there's a similar problem [here](https://github.com/noirbizarre/flask-restplus/issues/641), but it's an odd-ball solution. 
[...]

PR #747, opened 2019-11-01
This code addresses custom validation on per-field basis

andreixk:
The #716 issue is addressed in this PR. Each field can specify its own validate() method and raise an exception producing appropriate error message.


andreixk:
Just wondering if this has been reviewed or prioritized?

Issue #576, opened 2019-01-26
Upload File Example Broken


patientplatypus:
This code (https://flask-restplus.readthedocs.io/en/stable/parsing.html#file-upload):

```
from werkzeug.datastructures import FileStorage

upload_parser = api.parser()
upload_parser.add_argument('file', location='files',
                           type=FileStorage, required=True)


@api.route('/upload/')
@api.expect(upload_parser)
class Upload(Resource):
    def post(self):
        uploaded_file = args['file']  # This is FileStorage instance
        url = do_something_with_file(uploaded_file)
        return {'url': url}, 201
```

Doesn't compile. Maybe you need something like this `args = upload_parser.parse_args('file')`, but args is clearly not defined.

I've wasted too much time on this for me to consider using your library and it's obnoxious that you have code examples that don't work. Fix this.

Colin-b:
I'm not sure your tone is adequate here... There is indeed a mistake, thanks for pointing it out, you could submit a PR to fix it now that you found the issue.

Python is easy to access as a language but imho you should try to dig a bit more into the fundamentals before copy/pasting code.

Note that args could also be related to flask.request.args (even if using Flask restplus is the right way to do it in this case).

Lomomn:
Workaround:

```python
def post(self):
    if 'file' not in request.files:
        abort(400)
    
    file = request.files['file']
    if file.filename == '':
        abort(400)

    filename = secure_filename(file.filename)
    filepath = os.path.join(Config.UPLOAD_FOLDER, filename)
    file.save(filepath)

    return {'filepath': filename}
```

I can't get the parser to work with swagger yet, but simply sending the file works with postman and curl.
Not sure if this is what you were looking for, @patientplatypus

PR #749, opened 2019-11-23
Fix for #576 "Upload File Example Broken"

NAM-MAN:
see https://github.com/noirbizarre/flask-restplus/issues/576


andreixk:
@NAM-MAN can you add corresponding unit tests to ensure the changes are covered?

NAM-MAN:
@andreixk 
Looking at unittest changed my mind.
In this example, just parse like unittest.

Thank you!

PR #754, opened 2019-11-29
throw exception instead of abort for validate request

tannguyenit:
I think this library is for the API, so returning the error format should be user-editable, so I think that instead of aborting, we should throw the exception better, They can catch the case this and return with their own format, please see my pull request

### DIFF
--- a/flask_restplus/model.py
+++ b/flask_restplus/model.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
+from werkzeug import exceptions
 
 import copy
 import re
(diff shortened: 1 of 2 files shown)


j5awry:
please see our [contribution guide](https://github.com/noirbizarre/flask-restplus/blob/master/CONTRIBUTING.rst) for info on how to contribute. This PR lacks testing, has a failing build, and does not have an issue against it. Before a real review can begin, please address the issues with the PR.

Furthermore, reqparse has been slated for removal. Changing the behavior when there is discussion of removing it entirely doesn't make a lot of sense to me.

tannguyenit:
OK sir.

Issue #468, opened 2018-06-06
How to aggregate Swagger docs from multiple Blueprints?


Midnighter:
Maybe my design choices are flawed so feel free to correct me. My goal was to be able to assign URLs to several resources while keeping them in the same namespace. Thus I used blueprints which work great for me. My only problem is that now the docs are served at each individual blueprint URL rather than having an aggregated one. Pseudocode for my setup:

Resource `status.py`:
```python
from flask import Blueprint
from flask_restplus import Resource, Api

status = Blueprint("status", __name__)
api = Api(status, description="Query status.")

@api.route("/<string:uuid>")
@api.doc(...)
class Status(Resource):
    ...
```

In the main `app.py`:
```python
from flask import Flask

from .status import status

app = Flask(__name__)
app.register_blueprint(status, url_prefix="/status")
```

I have multiple such blueprints and would like to see all of their docs under one endpoint.

Fearxpc:
Hi,

you can try

status.py
`from flask_restplus import Namespace
api = Namespace('status', description='Status related stuff')

@api.route("/<string:uuid>")
@api.doc(...)
class Status(Resource):
    ...`

api.py
`blueprint = Blueprint('api', __name__, url_prefix='/myapi')
api = Api(blueprint,
    title='API',
    description='API'
)

from status import api as ns1
api.add_namespace(ns1)
`

app.py
`from flask import Flask

from .status import status

app = Flask(__name__)
app.register_blueprint(status, url_prefix="/status")
from api import blueprint as blueprint1
app.register_blueprint(blueprint1)`

ethe:
As @Fearxpc 's advice, we get a single blueprint app with multiple namespace. But if there are already multiple blueprints exist, how can we solve this problem?

Fearxpc:
@ethe 
You should be able to reigster multiple blueprints with `app.register_blueprint()` - that's why I used the var name blueprint1.

ethe:
@Fearxpc 

blueprint1 's endpoint is '\myapi' so I can  '/myapi' to show the blueprint1 swagger, that's right. But if I register another blueprint (maybe call the blueprint2 and it's endpoint is '/myapi2'), I can find the doc in '/myapi2'. But there are not a aggregation of blueprint1 and blueprint2, I can not find a page list all of the api from all blueprints. And the issue title is 'How to aggregate Swagger docs from multiple Blueprints?'.

Hope for your reply, thank you.

[...]

PR #785, opened 2020-03-11
added namespace example

james-powis:
The documentation currently does not define an example of a modularized restapi due to the overloading of variables in previous examples (ie api = Api(..) vs api = Namespace) this has caused some confusion (and took myself an embarrassing amount of time to finally find the nugget in https://github.com/noirbizarre/flask-restplus/blob/master/examples/zoo/cat.py which finally got everything working).

This PR aims to address the confusion encountered in #468 

However, it might be worthwhile to clean up variables names so we are:
`from .apis.namespace1 import namespace as ns1` instead of `from .apis.namespace1 import api as ns1`

I am not sure of the intention of the variable names, and perhaps there is some logic that escapes me, so I opted not to clean that up in this PR instead of choosing instead to provide a complete example that works.

### DIFF
--- a/doc/scaling.rst
+++ b/doc/scaling.rst
@@ -214,6 +214,20 @@ Given the previous layout, we can migrate it to the following directory structur
         ├── ...
         └── namespaceX.py
 
+Each `apis/namespaceX` wmodule will have the following pattern:
+
+.. code-block:: python
+
+    from flask_restplus import Namespace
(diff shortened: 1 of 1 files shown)


james-powis:
opened under https://github.com/python-restx/flask-restx/pull/78 as i did not see this had been forked.

